PR #2, opened 2018-10-31
Updated noble-mac dependency

churchs19:
Updated noble-mac dependency to [PR version that I submitted to noble-mac](https://github.com/Timeular/noble-mac/pull/8) that will work correctly on Linux/Windows. The recent change to noble-mac broke node-poweredup on any non-Mac OS.

I've tested this fix on Linux. The Noble library doesn't work with my Windows Bluetooth adapter, but it should work the same there. I don't have access to Mac OS to test, but there shouldn't be any issue with Mac as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -866,8 +866,8 @@
       }
     },
     "noble-mac": {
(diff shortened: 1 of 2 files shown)


nathankellenicki:
Thanks, I haven't had a chance to test on platforms other than macOS or Raspbian lately!

PR #3, opened 2018-11-15
Minor refactoring for NPM/TypeScript package compatibility

churchs19:
I moved the library exports into index.ts from poweredup.ts since TypeScript looks for an index.ts or index.d.ts as the root of the module when resolving modules imported through NPM. Following the existing instructions to 

`npm install node-poweredup --save`

and then referencing

`const PoweredUP = require("node-poweredup");`

in TypeScript results in a "Unable to find module 'node-poweredup'" error when running `tsc` despite the module being present in node_modules. By moving the exports to index.ts, I corrected this error and verified that the library works correctly against my Boost Movehub when importing as an npm package.

### DIFF
--- a/poweredup.ts
+++ b/poweredup.ts
@@ -16,7 +16,6 @@ import { EventEmitter } from "events";
 import Debug = require("debug");
 const debug = Debug("PoweredUP");
 import noble = require("noble-mac");
-import { start } from "repl";
 
 let ready = false;
 let wantScan = false;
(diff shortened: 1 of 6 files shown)


poweredup.ts
```diff
@@ -16,7 +16,6 @@ import { EventEmitter } from "events";
 import Debug = require("debug");
 const debug = Debug("PoweredUP");
 import noble = require("noble-mac");
-import { start } from "repl";
```
nathankellenicki: Thank you, this PR looks good! I hadn't even noticed that this had gotten in here. I assume my IDE inserted it at some point without me realising...

churchs19: Who knows? I just removed it because my IDE showed it wasn't actually being used anywhere.

PR #4, opened 2018-11-16
Extend LPF2Hub events with "colorAndDistance"

tomaszni:
Emitting both color and distance from Boost sensor at the same time, just like the sensor information is delivered.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -425,6 +425,7 @@ Emits when an attached motor or sensor is detached from the Hub.
     * ["button" (button, state)](#LPF2Hub+event_button)
     * ["distance" (port, distance)](#LPF2Hub+event_distance)
     * ["color" (port, color)](#LPF2Hub+event_color)
+    * ["colorAndDistance" (port, color, distance)](#LPF2Hub+event_colorAndDistance)
     * ["tilt" (port, x, y)](#LPF2Hub+event_tilt)
     * ["rotate" (port, rotation)](#LPF2Hub+event_rotate)
     * ["attach" (port, type)](#Hub+event_attach)
@@ -698,6 +699,19 @@ Emits when a color sensor is activated.
 | port | <code>string</code> |  |
 | color | <code>number</code> | A number representing one of the LED color consts. |
 
+<a name="LPF2Hub+event_colorAndDistance"></a>
+
+### "colorAndDistance" (port, color, distance)
+A combined color and distance event, emits when the sensor is activated.
+
+**Kind**: event emitted by [<code>BoostMoveHub</code>](#BoostMoveHub)  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| port | <code>string</code> |  |
+| color | <code>number</code> | A number representing one of the LED color consts. |
+| distance | <code>number</code> | Distance, in millimeters. |
+
 <a name="LPF2Hub+event_tilt"></a>
 
 ### "tilt" (port, x, y)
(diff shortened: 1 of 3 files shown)


nathankellenicki:
Thanks for the PR! This looks good, I'll test and merge.

May I ask what your use case is for requiring both in the same event, that can't be achieved by listening on both individual events?

tomaszni:
I used it for detecting colored bricks/tiles on LEGO train tracks - very much akin to your example ;)

In order to enhance detection (reduce the false positives rate) I utilized distance information to discern real tiles/bricks from background color (we ran the train on a red carpet) or erroneous reads (very often blue would register - probably as a reflection of the LEDs from the train track).

I started with using separate events, but keeping track of last seen color and then checking distance on next event was finicky, and introduced unnecessary complication to the code - having both pieces of data at the same time was the right way to go :)

As an interesting side note - I've managed to detect black tiles (despite no color / black being 0 value) because the distance was greater than usual (in my case it registered as >80mm), and to detect red tiles on red background (background carpet registered at least > 0, while the tile registered at < 0).

icedice:
> I used it for detecting colored bricks/tiles on LEGO train tracks - very much akin to your example ;)
> 
> In order to enhance detection (reduce the false positives rate) I utilized distance information to discern real tiles/bricks from background color (we ran the train on a red carpet) or erroneous reads (very often blue would register - probably as a reflection of the LEDs from the train track).
> 
> I started with using separate events, but keeping track of last seen color and then checking distance on next event was finicky, and introduced unnecessary complication to the code - having both pieces of data at the same time was the right way to go :)
> 
> As an interesting side note - I've managed to detect black tiles (despite no color / black being 0 value) because the distance was greater than usual (in my case it registered as >80mm), and to detect red tiles on red background (background carpet registered at least > 0, while the tile registered at < 0).

What trains are you using? Does the duplo trains actually have distance sensors? Im doing a swift app, using this lip as reference but I'm getting way to many false positives to use the detected tiles for anything...

tomaszni:
We are using PoweredUp trains (the yellow, passenger one). 
I've added LEGO Boost color and distance sensor to the train (connected to a second channel of the hub), looking down at the track (mounted through the chassis hole).

As far as false positives go - you need to look both at the color AND the distance reported (that was the whole point of the PR ;) By keeping only the correct color and distance combinations, you can filter out most of the false positives.

Let me give you a snippet of code that worked for me:
```
    let lastColor = -1;
    let lastTile = '';

    hub.on("colorAndDistance", async (port, color, distance) => {
        let tile = '';

        if (color == 0 && distance > 80)
            tile = 'BLACK';

        if (color == 9 && distance <= 0)
            tile = 'RED';
        if (color == 7 && distance <= 18)
            tile = 'YELLOW';
        if (color == 5 && distance <= 0) 
            tile = 'GREEN';
        if (color == 10 && distance <= 18)
            tile = 'WHITE';
        if (color == 3 && distance <= 10)
            tile = 'BLUE';

        if (tile && tile != lastTile) {
            debug('seen tile', tile);
            // do something with that knowledge here :)
        }

        if (color == lastColor) {
            if (tile)
                lastTile = tile;
            trace("same", port, color, '@', distance);
            return;
        }

        trace("diff", port, color, '@', distance);

        lastTile = tile;
        lastColor = color;
    });
```

Of course you would probably need to adjust the threshold levels to your case and environment, but the general idea holds.

Disclaimer: you should change the numerical color values to constants from the library, 'coz _everytime you hardcode values a kitten dies_ ;)

undera:
Just to share: From my experience with this color sensor, distance between sensor and target matters a lot. At some distance it detects color the best. So it makes sense to experiment with changing distance, if your mount allows that.
Also I used to accumulate several values of data within short interval, to sort out possible noise from consistent values.

icedice:
Thanks for the input.. actually I just found out that I can but the sensor in "C_TAG" mode which detects the action bricks perfectly... :)

nathankellenicki:
Are you talking about the Duplo train base in particular? Can I ask you to share how you achieved this? :)

icedice:
Yes this was speciffic for duplo train base... when I subscribe for notifications from the sensor I put it in mode 1 ( which is C_TAG mode) instead of mode 0 (which is color detect)

   //                 size, unused, cmd ,port - MODE - |-DELTA INTERVAL-|  subsribe
        cmd =  [0x0a, 0x00, 0x41, 0x12, 0x01, 0x01, 0x00, 0x00, 0x00, 0x01] // subscribe on color detect

PR #6, opened 2018-12-24
DuploTrainBaseSounds BREAK & STEAM added

techi602:
I have discovered new sounds for Duplo steam train model no. 10874

### DIFF
--- a/consts.ts
+++ b/consts.ts
@@ -54,9 +54,11 @@ export enum ButtonStates {
 
 
 export enum DuploTrainBaseSounds {
+    BREAK = 3,
     STATION_DEPARTURE = 5,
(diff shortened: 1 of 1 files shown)


nathankellenicki:
Thank you! Do you know if those work on the cargo train, and vice versa?

techi602:
Unfortunately I own only the steam train - all those constants work for steam train. However the wheel platform looks identical so I would assume that they also have identical microcontroller.

nathankellenicki:
Interesting - I wonder why I didn’t discover those other ones on my cargo train. I iterated through them all I thought! 

Nonetheless, thanks the the PR, I’ll gladly merge. Merry Christmas! :)

PR #8, opened 2018-12-26
Update duplotrainbase.ts

jankeymeulen:
Some sounds I discovered when playing with this.

### DIFF
--- a/duplotrainbase.ts
+++ b/duplotrainbase.ts
@@ -140,6 +140,13 @@ export class DuploTrainBase extends LPF2Hub {
      * @method DuploTrainBase#playSound
      * @param {number} sound A number representing one of the Train Base sound consts.
      * @returns {Promise} Resolved upon successful issuance of command.
+     *
+     * For the 10874 Duplo Train Set these sounds are
+     *  2 = brake
+     *  4 = chime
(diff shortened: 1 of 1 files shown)


nathankellenicki:
Hi, thanks!

The sounds are documented under consts inside consts.ts - https://github.com/nathankellenicki/node-poweredup/blob/master/consts.ts#L56-L62

You're right though in that the consts could be better surfaced in the documentation however. I'm going to expose these in the jsdocs like everything else after the holidays, but for now I'm happy to merge this. :)

Could you do a "npm run docs" to regenerate the readme with your change and commit it to this PR?

nathankellenicki:
Hi, as mentioned previously, I've added proper documentation for the various types and where they are used. See these links for an example for duploTrainBase.playSound and DuploTrainBaseSounds:

https://github.com/nathankellenicki/node-poweredup#DuploTrainBase+playSound
https://github.com/nathankellenicki/node-poweredup#DuploTrainBaseSound

Thanks for prompting me to add this! I'm closing PR as no longer necessary.

jankeymeulen:
Great, thanks a lot! I'm absolutely terrible in (Node.)js so thanks for updating the docs!

PR #12, opened 2019-01-16
remove colorAndDistance event from DuploTrainBase docs

firien:
I don't think the Duplo train has a distance sensor.


nathankellenicki:
Good spot - I'd already excluded "distance" from the DuploTrainBase, but hadn't thought about the "colorAndDistance" event introduced later. Thanks!

PR #17, opened 2019-02-04
Fixed broken device type query

davidcolombogit:
Replaced "Devices.LED_LIGHTS" with "DeviceType.LED_LIGHTS" 
Because "Devices.LED_LIGHTS" is undefined (maybe deprecated)


nathankellenicki:
Thank you! Yes, I renamed/deprecated the old way a while back and clearly forgot to update a few examples. 👍

PR #31, opened 2019-10-13
Updating noble dependency to @abandonware/noble for Node >= 10.0 support

churchs19:
Noble hasn't been maintained in over 2 years and no longer builds on Node.js >= 10.0. The [Abandonware Project](https://abandonware.github.io/) has taken on maintenance and updated dependencies to allow Noble to build and run with no breaking changes on Node.js 10+. 

As a note, the new abandonware-noble.d.ts is simply the same definition file from @types/noble wrapped for the @abandonware/noble module since there aren't type definitions for the abandonware fork yet.


nathankellenicki:
Thanks, this is excellent! I think I can probably remove the `noble-mac` dependency too as a result.

churchs19:
I think you're right as I look into that more. That would simplify the dependency chain as well.

PR #33, opened 2019-11-02
allow RGB led control for the duplo train hub

nutki:
Duplo train LED has both indexed color and rgb modes just like other hubs, so it can be controlled with the same code.


nathankellenicki:
Thank you, this looks great! Code de-duplication for the win.

PR #35, opened 2019-11-02
Fix and simplify tilt sensor reading

nutki:
WeDo tilt sensor in Mode 0 and Boost Move tilt sensor in mode 4 both return signed 8 bit integers. So they can be directly read as such.
The existing code was causing 2 issues:
- WeDo tilt X axis value was always positive (in both directions)
- Boost move mode 4 range is -128 to 127, so some negative values were reported as positive in range 128 - 160
Also the sign of the raw read was sometimes reversed, not sure if this was intentional. I undone it, but could be reverted to maintain backwards compatibility.

### DIFF
--- a/src/lpf2hub.ts
+++ b/src/lpf2hub.ts
@@ -496,8 +496,8 @@ export class LPF2Hub extends Hub {
                     break;
                 }
                 case Consts.DeviceType.WEDO2_TILT: {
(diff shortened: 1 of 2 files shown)


nathankellenicki:
Good catches. I hadn't noticed it being reversed. I'd rather introduce the fix (and reversal), I'll bump the semver appropriately.

PR #41, opened 2019-11-05
LPF2 firmware versions should be decoded with BCD

nutki:
As per the official protocol specification all 4 byte version numbers are using BCD number encoding: https://lego.github.io/lego-ble-wireless-protocol-docs/index.html#version-number-encoding

### DIFF
--- a/src/boostmovehub.ts
+++ b/src/boostmovehub.ts
@@ -304,7 +304,7 @@ export class BoostMoveHub extends LPF2Hub {
 
 
     protected _checkFirmware (version: string) {
-        if (compareVersion("2.0.00.0023", version) === 1) {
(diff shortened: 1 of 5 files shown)


nathankellenicki:
Thanks, good catch. I ummed and ahed about storing the parts as a string rather than individually, but this way will be more consistent. Merged.

nutki:
I did not have specific reason to change the version values to be stored as unstructured string on its own, but switching to BCD tipped the scale. The easiest way to convert from BCD is to print as hex and parse it as integers. But then the next operation would be to turn it into string again. Also the fact that in the spec the version is treated as a 32bit number and not 4 distinct values makes the conversion matching the spec more.

PR #42, opened 2019-11-09
Remove sources from npmignore

aileo:
It allows to import the sources to be bundled with main app if needed

E.g:
```js
import { PoweredUp } from 'node-poweredup/src/poweredup-browser';

const poweredUP = new PoweredUP();
```
Then the lib is not exposed to the `window` scope and bundled with the app source.


nathankellenicki:
Good idea, thanks! Will merge.

PR #44, opened 2019-11-12
Voltage and current readout improvements

nutki:
This reorganizes the LPF2 hub voltage and current readout code adding some improvements:
- voltage readout in Duplo Train base added
- voltage readout in PUP Remote was overridden by RSSI value

### DIFF
--- a/src/controlplushub.ts
+++ b/src/controlplushub.ts
@@ -32,7 +32,6 @@ export class ControlPlusHub extends LPF2Hub {
 
     protected _currentPort = 0x3b;
     protected _voltagePort = 0x3c;
-    protected _voltageMaxV = 9.615;
     protected _voltageMaxRaw = 4095;
 
     constructor (device: IBLEDevice, autoSubscribe: boolean = true) {
--- a/src/controlplushub.ts
+++ b/src/controlplushub.ts
@@ -30,6 +30,10 @@ export class ControlPlusHub extends LPF2Hub {
         );
     }
 
+    protected _currentPort = 0x3b;
+    protected _voltagePort = 0x3c;
+    protected _voltageMaxV = 9.612;
+    protected _voltageMaxRaw = 4095;
 
     constructor (device: IBLEDevice, autoSubscribe: boolean = true) {
         super(device, autoSubscribe);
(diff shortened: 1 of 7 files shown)


src/hub.ts
```diff
@@ -31,7 +31,6 @@ export class Hub extends EventEmitter {
     protected _current: number = 0;
 
     protected _bleDevice: IBLEDevice;
-    private _rssi: number = -100;
```
nutki: I can reintroduce RSSI in this or a followup PR. In the current state the `_rssi` was never changed from the initial.

src/controlplushub.ts
```diff
@@ -30,6 +30,10 @@ export class ControlPlusHub extends LPF2Hub {
         );
     }
 
+    protected _currentPort = 0x3b;
+    protected _voltagePort = 0x3c;
+    protected _voltageMaxV = 9.615;
```
nutki: I copied this value `9615 / 1000` from the existing code (same with `9620 / 1000` for the PUP hub). However the port information data reports `9600 / 1000` in both cases (like the control plus hub). Can you point me to the source of these values (are they coming from an earlier FW)?

src/hub.ts
nathankellenicki: Oh that's definitely a bug. RSSI should totally be accessible. But if it's not working at the moment it's fine to remove it in this PR - just need to remember to fix it.

src/controlplushub.ts
nathankellenicki: So I think the 9614/9620 value came around before we knew how to do port information requests (ie. Before Lego released the LWP document). It was determined by someone actually measuring the voltage. So, pretty close!

I'd say the port information request is probably the source of truth though.

nutki: Well, this is a impressive result indeed. I think in this case I may unify these values to use the same `_voltageMax` to reduce the code size. The difference in values reported will be less than 1% anyway.

src/hub.ts
nutki: I can follow up with the RSSI readout PR. In the previous code the branch that supposed to read the value was missing: https://github.com/nathankellenicki/node-poweredup/blob/0fbd27d41ff49b435ae7df260bd06200b13c8433/src/lpf2hub.ts#L403

PR #45, opened 2019-11-12
 Extended port information requests

nutki:
This extends the information obtained with port mode info requests and disables them by default (to reduce unnecessary radio chatter).
New information includes:
- Identifying all know device types by name
- Port mode combinations
- FW/HW versions
- value format

### DIFF
--- a/src/consts.ts
+++ b/src/consts.ts
@@ -29,7 +29,10 @@ export const HubTypeNames = HubType;
  * @property {number} BASIC_MOTOR 1
  * @property {number} TRAIN_MOTOR 2
  * @property {number} LED_LIGHTS 8
- * @property {number} BOOST_LED 22
+ * @property {number} VOLTAGE 20
+ * @property {number} CURRENT 21
+ * @property {number} PIEZO_TONE 22
+ * @property {number} RGB_LIGHT 23
  * @property {number} WEDO2_TILT 34
  * @property {number} WEDO2_DISTANCE 35
  * @property {number} BOOST_DISTANCE 37
@@ -43,6 +46,7 @@ export const HubTypeNames = HubType;
  * @property {number} CONTROL_PLUS_LARGE_MOTOR 46
  * @property {number} CONTROL_PLUS_XLARGE_MOTOR 47
  * @property {number} POWERED_UP_REMOTE_BUTTON 55
+ * @property {number} RSSI 56
  * @property {number} CONTROL_PLUS_ACCELEROMETER 58
  * @property {number} CONTROL_PLUS_TILT 59
  */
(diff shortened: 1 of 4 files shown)


src/consts.ts
```diff
@@ -20,11 +20,7 @@ export enum HubType {
 
 
 // tslint:disable-next-line
-export let HubTypeNames = Object.keys(HubType).reduce((result: {[hubType: string]: string}, item) => {
-    // @ts-ignore
-    result[HubType[item]] = item;
-    return result;
-}, {});
+export const HubTypeNames = HubType;
```
nutki: TS enums with number values include the reverse mapping already, so the code above produced HubTypeNames which was almost identical in content to HubType enum, but the old `DeviceTypeNames` where causing problems with TS typing (previously they were all unused)

src/consts.ts
```diff
@@ -47,6 +46,7 @@ export let HubTypeNames = Object.keys(HubType).reduce((result: {[hubType: string
  * @property {number} CONTROL_PLUS_LARGE_MOTOR 46
  * @property {number} CONTROL_PLUS_XLARGE_MOTOR 47
  * @property {number} POWERED_UP_REMOTE_BUTTON 55
+ * @property {number} RSSI 56
  * @property {number} CONTROL_PLUS_ACCELEROMETER 58
  * @property {number} CONTROL_PLUS_TILT 59
```
nutki: I did not include here the mystery device type 66 which is an interval device in the Boost Hub at port 0x46. Do you have any idea what that is?

src/hub.ts
```diff
@@ -372,15 +372,6 @@ export class Hub extends EventEmitter {
         return port;
     }
 
-
-    protected _lpad (str: string, length: number) {
```
nutki: This function is unused since the version parsing changes, but I used it in my fork for the previous version of `toBin` and `toHex`.

src/consts.ts
nathankellenicki: I don't - but I've sent off a couple of emails to ask. I think for the moment lets leave it off.

nutki: I tested this on my own and this port seems to return 0 as a value in all 3 modes, so probably not interesting anyway.

src/lpf2hub.ts
```diff
@@ -20,6 +21,8 @@ export class LPF2Hub extends Hub {
         return [t[0], t[1], t.substring(2, 4), t.substring(4)].join(".");
     }
 
+    public sendPortInformationRequests: boolean = false;
```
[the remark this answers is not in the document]
nutki: Yes, this is a good idea to make it consistent this way. I am not familiar with the debug module so was not sure if this is possible. I will update this.

On the other hand I plan to follow up at some point with a change to report this information in a structured way with event emitters and a standalone command line tool to output full hub information.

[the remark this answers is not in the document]
nathankellenicki: Makes sense, and sounds like a good tool. I think until the tool is implemented we should query the debug setting, and we can revisit later with a different check, depending on how it needs to be set.

PR #47, opened 2019-11-13
Add mac address to hub/lpf2hub

aileo:
- should close nathankellenicki/node-poweredup/issues/46

### DIFF
--- a/src/hub.ts
+++ b/src/hub.ts
@@ -26,7 +26,7 @@ export class Hub extends EventEmitter {
     protected _name: string = "";
     protected _firmwareVersion: string = "0.0.00.0000";
     protected _hardwareVersion: string = "0.0.00.0000";
-    protected _macAddress: string = "00:00:00:00:00:00";
+    protected _primaryMACAddress: string = "00:00:00:00:00:00";
     protected _batteryLevel: number = 100;
     protected _voltage: number = 0;
     protected _current: number = 0;
@@ -75,10 +75,10 @@ export class Hub extends EventEmitter {
 
     /**
      * @readonly
-     * @property {string} macAddress Primary mac address of the hub
+     * @property {string} primaryMACAddress Primary MAC address of the hub
      */
-    public get macAddress () {
-        return this._macAddress;
+    public get primaryMACAddress () {
+        return this._primaryMACAddress;
     }
 
 
--- a/src/hub.ts
+++ b/src/hub.ts
@@ -25,6 +25,7 @@ export class Hub extends EventEmitter {
 
     protected _name: string = "";
     protected _firmwareVersion: string = "0.0.00.0000";
+    protected _macAddress: string = "00:00:00:00:00:00";
     protected _batteryLevel: number = 100;
     protected _voltage: number = 0;
     protected _current: number = 0;
@@ -60,6 +61,15 @@ export class Hub extends EventEmitter {
      */
     public get firmwareVersion () {
         return this._firmwareVersion;
+    }
+
+
+    /**
+     * @readonly
+     * @property {string} macAddress Primary mac address of the hub
+     */
+    public get macAddress () {
+        return this._macAddress;
     }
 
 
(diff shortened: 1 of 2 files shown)


src/lpf2hub.ts
```diff
@@ -20,6 +20,17 @@ export class LPF2Hub extends Hub {
...
 
+    private static decodeMacAddress(v: Uint8Array) {
+        return [
+            ("00" + v[0].toString(16)).slice(-2),
+            ("00" + v[1].toString(16)).slice(-2),
+            ("00" + v[2].toString(16)).slice(-2),
+            ("00" + v[3].toString(16)).slice(-2),
+            ("00" + v[4].toString(16)).slice(-2),
+            ("00" + v[5].toString(16)).slice(-2),
+        ].join(":");
```
aileo: Not the ideal way to parse it, I tried `return v.map((n) => ("00" + n.toString(16)).slice(-2)).join(":");` but TS would not compile.

dlech: try

    return Array.from(v).map((n) => ("00" + n.toString(16)).slice(-2)).join(":");

map of Uint8Array returns a typed array, so can't handle the strings.

src/hub.ts
```diff
@@ -63,6 +64,15 @@ export class Hub extends EventEmitter {
     }
 
 
+    /**
+     * @readonly
+     * @property {string} macAddress Primary mac address of the hub
+     */
+    public get macAddress () {
```
dlech: Technically speaking this is not a MAC address, it is a Bluetooth address (they just happen to have the same format).

src/hub.ts
```diff
@@ -64,6 +65,15 @@ export class Hub extends EventEmitter {
     }
 
 
+    /**
+     * @readonly
+     * @property {string} firmwareVersion Hardware version of the hub
```
dlech: is this supposed to say `hardwareVersion` instead of `firmwareVersion`?

src/hub.ts
aileo: As the LWP refers to it as "Primary MAC Address", I was going for the same terminology but maybe simply "address" would be more technically accurate.

dlech: Most people will never know the difference anyway :smile:

src/lpf2hub.ts
```diff
@@ -20,6 +20,10 @@ export class LPF2Hub extends Hub {
         return [t[0], t[1], t.substring(2, 4), t.substring(4)].join(".");
     }
 
+    private static decodeMacAddress(v: Uint8Array) {
+        return Array.from(v).map((n) => ("00" + n.toString(16)).slice(-2)).join(":");
```
nutki: This is simpler now with ECMASctipt 2017. I also happened to add an utility function `toHex` which does exactly ` n.toString(16).padStart(2, "0")` in my own PR.

```suggestion
        return Array.from(v).map((n) => n.toString(16).padStart(2, "0")).join(":");
```

aileo: I will keep in mind to use your `toHex` as soon as it get merged.

src/hub.ts
nathankellenicki: Small niggle triggering my OCD - can we standardise on "MAC" address (uppercase), as used below? :)

nathankellenicki: Regarding MAC address vs just address, I guess it makes most sense to follow Lego's own terminology, since this is reportable only through their protocol. I guess it would cause less confusion if it was consistent?

src/hub.ts
```diff
@@ -73,6 +74,15 @@ export class Hub extends EventEmitter {
     }
 
 
+    /**
+     * @readonly
+     * @property {string} macAddress Primary mac address of the hub
```
nathankellenicki: Capitalisation of "MAC address"

src/hub.ts
aileo: Then I would go for `primaryMACAddress` as secondary one exists in the protocol. What do you think ?

nathankellenicki: I like it, that sounds good to me!

src/lpf2hub.ts
nathankellenicki: I merged @nutki's PR, feel free to use `toHex` now :)

aileo:
> LGTM other than the MAC address comment! Thanks!

Everything should be ok now :)

nathankellenicki:
Thanks!

PR #48, opened 2019-11-13
Add PoweredUP#getConnectedHubByPrimaryMACAddress

aileo:
Predictable pull request following #47 

Allow to retrieve hub by its primary MAC address but:
- it assumes those addresses are really unique (they should be)
- if not it returns the first hub found matching provided address

### DIFF
--- a/src/poweredup-browser.ts
+++ b/src/poweredup-browser.ts
@@ -97,6 +97,17 @@ export class PoweredUP extends EventEmitter {
 
 
     /**
+     * Retrieve a Powered UP Hub by primary MAC address.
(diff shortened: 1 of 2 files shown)


aileo:
I don't know if it really belongs to the lib as it is some `getConnectedHubs` with a filter...

On the other hand, `getConnectedHubsByName` exists so why not, but it opens the door to every fantasy:
- `getConnectedHubsByType`
- `getConnectedHubsByDeviceMapping`

nathankellenicki:
I think this is a fair addition. :) I can see your point, but I think looking up a connected hub by some kind of identifier is a useful feature. I'd argue the rest of the properties don't count as identifiers per se.

PR #51, opened 2019-11-15
hubinfo tool

nutki:
Added a tool which can display information about discovered hubs (including detailed port information).

Example output
```
e07dea09c7a1 firmware v1.2.00.0000 hardware v1.0.00.0000 DUPLO_TRAIN_HUB (Train Base)
90842b0e7cad firmware v1.1.02.0000 hardware v0.1.00.0000 POWERED_UP_HUB (HUB NO.4)
0081f9059568 firmware v1.0.00.0004 hardware v1.0.00.0000 POWERED_UP_REMOTE ()
001653b4d6a9 firmware v2.0.00.0017 hardware v0.4.00.0000 BOOST_MOVE_HUB (LEGO Move Hub)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,5 +32,8 @@
     "typescript": "^3.7.2",
     "webpack": "^4.41.2",
     "webpack-cli": "^3.3.10"
+  },
+  "bin": {
+    "lpf2-hubinfo": "./bin/hubinfo"
   }
 }
--- a/src/hubinfo.ts
+++ b/src/hubinfo.ts
@@ -73,7 +73,7 @@ export function main(argv: string[]) {
         });
         await hub.sleep(2000);
         const typeName = Consts.HubTypeNames[hub.type] || `unknown (${toHex(hub.type)})`;
-        console.log(`${hub.uuid} ${hub.primaryMACAddress} firmware v${hub.firmwareVersion} hardware v${hub.hardwareVersion} ${typeName} (${hubName})`);
+        console.log(`${hub.uuid} firmware v${hub.firmwareVersion} hardware v${hub.hardwareVersion} ${typeName} (${hubName})`);
         if (hub instanceof LPF2Hub && !portInfo) {
             await hub.shutdown();
         }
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -32,5 +32,8 @@
     "typescript": "^3.7.2",
     "webpack": "^4.41.2",
     "webpack-cli": "^3.3.10"
+  },
+  "bin": {
+    "lpf2-hubinfo": "./bin/hubinfo"
```
nathankellenicki: For consistency with this library name, could we call it `poweredup-hubinfo`? I'm not sure LPF2 is the right name to use given that its Lego internal only, and most people know the technology as Powered UP.

src/hubinfo.ts
```diff
@@ -0,0 +1,85 @@
...
+            case "-p":
+            case "--portinfo":
+                portInfo = true;
+                break;
+            case "-q":
+            case "--quiet":
+                quiet = true;
+                break;
+            default:
+                console.log("Usage: lpf2-hubinfo [-p|--portinfo] [-q|--quiet]");
```
nathankellenicki: The process never quits, likely as Noble is still running in the background. Can you add something like `process.exit(0);`?

src/hubinfo.ts
```diff
@@ -0,0 +1,85 @@
...
+                case 0x80: // Value Format
+                    console.log(`${prefix}, Value ${info.numValues} x ${info.dataType}, Decimal format ${info.decimalFormat}`);
+                    break;
+            }
+        });
+        await hub.sleep(2000);
+        const typeName = Consts.HubTypeNames[hub.type] || `unknown (${toHex(hub.type)})`;
+        console.log(`${hub.uuid} firmware v${hub.firmwareVersion} hardware v${hub.hardwareVersion} ${typeName} (${hubName})`);
+        if (hub instanceof LPF2Hub && !portInfo) {
+            await hub.shutdown();
```
nathankellenicki: Could we add another `process.exit(0);` here so that the process terminates after shutting down the hub?

nathankellenicki (changes requested):
Thanks for this, this looks very useful! I added a few small comments.

I noticed that the changes for port mode info were added to `lpf2hub.ts` which seems correct, but did you test this with a WeDo 2.0 Smart Hub, to check its handled gracefully?

If you don't have one to hand I can test myself when I get back home in a couple of weeks. But I won't block the merging of this on that.

PR #57, opened 2019-11-24
Control+ hub improvements

nutki:
I got my hands on a Control+ Hub and tested some build in sensor readings.


src/consts.ts
```diff
@@ -71,10 +71,13 @@ export enum DeviceType {
     DUPLO_TRAIN_BASE_SPEEDOMETER = 44,
     CONTROL_PLUS_LARGE_MOTOR = 46,
     CONTROL_PLUS_XLARGE_MOTOR = 47,
+    CONTROL_PLUS_GEST = 54,
```
nutki: I checked the values returned by this sensor. This seems to be a value in the range of 0-4 generated from the motion sensors. Most likely GEST stands for "gesture". Shaking gives 3. Tossing up - 4 when raising and 1 or 2 when landing. 0 when no movement.

src/controlplushub.ts
```diff
@@ -31,8 +31,10 @@ export class ControlPlusHub extends LPF2Hub {
     }
 
     protected _currentPort = 0x3b;
+    protected _currentMaxMA = 4175;
```
nutki: This is the value reported by Hub port info. I tried to verify it and it is probably better than the original value of 2444.
I used http://www.philohome.com/motors/motorcomp.htm as the reference with different motors with no load but it was inconclusive. http://www.philohome.com/wedo2reverse/consumption.htm does not include Control+ hub unfortunately.

src/lpf2hub.ts
```diff
@@ -548,12 +548,27 @@ export class LPF2Hub extends Hub {
...
+                     * @event LPF2Hub#gyro
+                     * @param {string} port
+                     * @param {number} x
+                     * @param {number} y
+                     * @param {number} z
+                     */
+                    this.emit("gyro", "GYRO", gyroX, gyroY, gyroZ);
+                    break;
+                }
                 case Consts.DeviceType.CONTROL_PLUS_ACCELEROMETER: {
```
nutki: This was previously reporting gyroscope data.

src/lpf2hub.ts
```diff
@@ -548,12 +548,27 @@ export class LPF2Hub extends Hub {
...
-                    const accelX = Math.round((data.readInt16LE(4) / 28571) * 2000);
-                    const accelY = Math.round((data.readInt16LE(6) / 28571) * 2000);
-                    const accelZ = Math.round((data.readInt16LE(8) / 28571) * 2000);
+                    const accelX = Math.round(data.readInt16LE(4) / 4.096);
+                    const accelY = Math.round(data.readInt16LE(6) / 4.096);
+                    const accelZ = Math.round(data.readInt16LE(8) / 4.096);
```
nutki: The scaling values for acceleration as reported by port info, verified to be correct.

src/lpf2hub.ts
```diff
@@ -548,12 +548,27 @@ export class LPF2Hub extends Hub {
                     this.emit("tilt", "TILT", this._lastTiltX, this._lastTiltY, this._lastTiltZ);
                     break;
                 }
+                case Consts.DeviceType.CONTROL_PLUS_GYRO: {
+                    const gyroX = Math.round(data.readInt16LE(4) * 7 / 400);
```
nutki: The reported scaling value ` 2000 / 28571.42` (which is actually equal to 7 / 100) seemed to produce values too large, so I calibrated it using a 33/45 rpm (200/270dps) record player which showed the values where exactly 4 times off (hence `7 / 400`).
Also the reading seems to show non-calibrated values, maybe the library could provide calibration for that.

src/controlplushub.ts
nutki: I confirmed the value of `4715` is using a multimeter.

nathankellenicki:
Thanks for this PR, great work! I like that this doesn't break the existing accel interface, even though the values are likely to be different, so it probably warrants a major version bump.

PR #56, opened 2019-11-21
Attempt to read differents device modes

aileo:
**This is not ready to be merged**, working on #55 took me further than I expected and I ended trying to read every input mode from every device I have.

Now I need help/advice/review to check what I did wrong and test devices I don't have (WeDo2 hub and accessories, Duplo train base)

To handle different data from different mode I added several events:

- BOOST_DISTANCE
  - count
  - reflectivity
  - luminosity
  - rgb
- Motors
  - power
  - speed
  - absolutePosition
- BOOST_TILT
  - angle
  - orientation
  - impact
  - accel

---

Using the debug output I summed up mode numbers and names for each hub.
The bad news is: the same device on different hubs does not expose exactly the same modes.

### Color and distance

| HUB              | 0x00  | 0x01  | 0x02  | 0x03  | 0x04  | 0x05  | 0x06  | 0x07  | 0x08   | 0x09  | 0x0a  |
| ---------------- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ------ | ----- | ----- |
| WEDO2_SMART_HUB  | ?     | ?     | ?     | ?     | ?     | ?     | ?     | ?     | ?      | ?     | ?     |
| BOOST_MOVE_HUB   | COLOR | PROX  | COUNT | REFLT | AMBI  | COL_O | RGB_I | IR_Tx | SPEC_1 | DEBUG | CALIB |
| POWERED_UP_HUB   | COLOR | PROX  | COUNT | REFLT | AMBI  | COL_O | RGB_I | IR_Tx | SPEC_1 | DEBUG | CALIB |
| CONTROL_PLUS_HUB | COLOR | PROX  | COUNT | REFLT | AMBI  | COL_O | RGB_I | IR_Tx | SPEC_1 | DEBUG | CALIB |

### Motors

**BASIC_MOTOR**

| HUB              | 0x00        |
| ---------------- | ----------- |
| WEDO2_SMART_HUB  | ?           |
| BOOST_MOVE_HUB   | LPF2-MMOTOR |
| POWERED_UP_HUB   | LPF2-MMOTOR |
| CONTROL_PLUS_HUB | LPF2-MMOTOR |

**BOOST_TACHO_MOTOR**

| HUB              | 0x00  | 0x01  | 0x02| 0x03 |
| ---------------- | ----- | ----- | --- | ---- |
| WEDO2_SMART_HUB  | ?     | ?     | ?   | ?    |
[...]

Issue #49, opened 2019-11-14
Discussion surrounding a new paradigm for talking to attached devices


nathankellenicki:
Something which has been on my mind for a while (which recently came up in a not-entirely-unrelated issue) has been a new paradigm for talking to devices attached to hubs.

Problem Space
---

Here are a few examples of the current design, and where it falls short.

```js
hub.setMotorSpeed("A", 50);
```

This method allows you to start running the motor attached to port "A" at a speed 50 constantly. It does no checking against the type of device actually attached to the port, and therefore may or may not work. For example, it will work just fine with an attached light (setting the brightness to half), but will fail against a color sensor.

If device type checking would be added, it would be a lengthy list of compatible devices, which could potentially get out of date as new devices are added.

Similarly:

```js
hub.setLightBrightness("A", 50);
```

Pretty much does the same as the above, but with less logic. For example `setMotorSpeed()` changes the underlying command depending on the attached motor and hub combination. However this doesn't - so it may set a motor running at half speed, it may not.

This method is used for rotating a tacho motor to a specified angle:

```js
hub.setMotorAngle("A", 50, 100);
```

It won't work against non-tacho motors, so a device type check is done at the start of the method (see above comment). As new devices are released with this capability, the method needs to be changed to add more devices to the device check. A similar problem exists with `setAbsoluteSpeed()`.

Another example is for setting the color of the LED on the hub. This is done like so:

```js
hub.setLEDColor(Consts.Color.PINK);
```

However the color of the LED on attached color sensor can also be changed, which this library doesn't support at the moment. Were it to support it, what would the solution be? Do we add an optional port argument to the method? Or do we create an entirely new method named something along the lines of `setLEDColorOnColorSensor()`? Neither seem ideal.

Current Proposal
---

Having thought about this for a little while, an approach that keeps coming to me is to create an individual class for each device, such as `SimpleMotor`, `TrainMotor`, `BoostMotor`, `Light`, `DistanceSensor`, `TiltSensor`, etc.

When one is attached to a hub, an "attach" event is emitted as normal, however instead of having the type id as a parameter, it would contain an instance of the device class, like so:

```js
hub.on("attach", (port, device) => {
    console.log(`Device attached - port ${port}, type ${device.type}`);
    if (device instanceof Light) {
        const light = device;
        light.setBrightness(50); // Set half brightness
    } else if (device instanceof BoostMotor) {
        const motor = device;
[...]

PR #60, opened 2019-12-04
Experiment: read sensor data with mode informations

aileo:
**Once again this is not meant to be merged as it breaks everything.**

Since the LWP3 exposes mode information with name and data format for parsing, I tried to implement some kind of "auto parsing":
- it parse data using mode information
- emit a `sensor` event with mode name and raw parsed data array (which allow to listen to events not handled by the lib)
- emit *almost* the same events the library already have

BUT: It does not work very well as I don't know how to get the mode information before the `attach` event is emited. As is, it just ignore sensor events until the mode information got collected...

If anyone see value in this and wants to help, I would be very happy to know if there is a way to make it reliable.

### DIFF
--- a/src/hub.ts
+++ b/src/hub.ts
@@ -161,9 +161,27 @@ export class Hub extends EventEmitter {
             if (mode !== undefined) {
                 newMode = mode;
             }
+            this._ports[port].mode = newMode;
             this._activatePortDevice(this._portLookup(port).value, this._portLookup(port).type, newMode, 0x00, () => {
                 return resolve();
             });
+        });
+    }
+
+    public subscribeByModeName (port: string, mode: string = "") {
+        const { value, type, modes } = this._portLookup(port);
+
+        return new Promise((resolve, reject) => {
+            this._activatePortDevice(
+                value,
+                type,
+                modes.reduce(
+                    (newMode, definition, index) => definition.name === mode ? index : newMode,
+                    this._getModeForDeviceType(type)
+                ),
+                0x00,
+                () => resolve()
+            );
         });
     }
 
@@ -176,6 +194,7 @@ export class Hub extends EventEmitter {
     public unsubscribe (port: string) {
         return new Promise((resolve, reject) => {
             const mode = this._getModeForDeviceType(this._portLookup(port).type);
+            this._ports[port].mode = null;
             this._deactivatePortDevice(this._portLookup(port).value, this._portLookup(port).type, mode, 0x00, () => {
                 return resolve();
             });
@@ -268,7 +287,7 @@ export class Hub extends EventEmitter {
         if (port.connected) {
             port.type = type;
             if (this.autoSubscribe) {
-                this._activatePortDevice(port.value, type, this._getModeForDeviceType(type), 0x00);
+                this.subscribe(port.id, this._getModeForDeviceType(type));
                 /**
                  * Emits when a motor or sensor is attached to the Hub.
                  * @event Hub#attach
--- a/src/hub.ts
+++ b/src/hub.ts
@@ -297,6 +297,7 @@ export class Hub extends EventEmitter {
             }
         } else {
             port.type = Consts.DeviceType.UNKNOWN;
+            port.modes = [];
             debug(`Port ${port.id} disconnected`);
             /**
              * Emits when an attached motor or sensor is detached from the Hub.
--- a/src/hub.ts
+++ b/src/hub.ts
@@ -168,21 +168,20 @@ export class Hub extends EventEmitter {
         });
     }
 
-    public subscribeByModeName (port: string, mode: string = "") {
-        const { value, type, modes } = this._portLookup(port);
+    public subscribeByModeName (port: string, mode?: string) {
+        const { modes } = this._portLookup(port);
+        const modeCode = modes.reduce(
+            (foundMode: number | null, definition, index) => {
+                return definition.name === mode ? index : foundMode;
+            },
+            null
+        );
 
-        return new Promise((resolve, reject) => {
-            this._activatePortDevice(
-                value,
-                type,
-                modes.reduce(
-                    (newMode, definition, index) => definition.name === mode ? index : newMode,
-                    this._getModeForDeviceType(type)
-                ),
-                0x00,
-                () => resolve()
-            );
-        });
+        if (!modeCode) {
+            throw new Error('Unknown device mode');
+        } else {
+            return this.subscribe(port, modeCode);
+        }
     }
 
     /**
--- a/src/lpf2hub.ts
+++ b/src/lpf2hub.ts
@@ -380,6 +380,7 @@ export class LPF2Hub extends Hub {
                 modeInfoDebug(`Port ${portHex}, mode ${mode}, name ${name}`);
                 break;
             case 0x01: // RAW Range
+                port.modes[mode].rawMin = data.readFloatLE(6);
                 modeInfoDebug(`Port ${portHex}, mode ${mode}, RAW min ${data.readFloatLE(6)}, max ${data.readFloatLE(10)}`);
                 break;
             case 0x02: // PCT Range
@@ -456,17 +457,18 @@ export class LPF2Hub extends Hub {
         if (port && port.connected && port.mode && port.modes[port.mode]) {
             const mode = port.modes[port.mode];
             const values = [];
+            const signed = mode.rawMin < 0;
 
             for (let index = 4; index < data.length; index += Consts.VALUE_SIZE[mode.valueType]) {
                 switch (mode.valueType) {
                     case Consts.ValueType.Int8:
-                        values.push(data.readInt8(index));
+                        values.push(signed ? data.readInt8(index) : data.readUInt8(index));
                         break;
                     case Consts.ValueType.Int16:
-                        values.push(data.readInt16LE(index));
+                        values.push(signed ? data.readInt16LE(index) : data.readUInt16LE(index));
                         break;
                     case Consts.ValueType.Int32:
-                        values.push(data.readInt32LE(index));
+                        values.push(signed ? data.readInt32LE(index) : data.readUInt32LE(index));
                         break;
                     case Consts.ValueType.Float:
                         values.push(data.readFloatLE(index));
--- a/src/lpf2hub.ts
+++ b/src/lpf2hub.ts
@@ -1,7 +1,7 @@
 import { Peripheral } from "@abandonware/noble";
 
 import { Hub } from "./hub";
-import { Port } from "./port";
+import { IPortMode, Port } from "./port";
 
 import * as Consts from "./consts";
 import { toBin, toHex } from "./utils";
@@ -294,8 +294,8 @@ export class LPF2Hub extends Hub {
             const hwVersion = LPF2Hub.decodeVersion(data.readInt32LE(7));
             const swVersion = LPF2Hub.decodeVersion(data.readInt32LE(11));
             modeInfoDebug(`Port ${toHex(data[3])}, hardware version ${hwVersion}, software version ${swVersion}`);
-            this._sendPortInformationRequest(data[3]);
         }
+        this._sendPortInformationRequest(data[3]);
 
         if (!port) {
             if (data[4] === 0x02) {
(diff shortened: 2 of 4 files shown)


src/hub.ts
```diff
@@ -268,7 +287,7 @@ export class Hub extends EventEmitter {
         if (port.connected) {
             port.type = type;
             if (this.autoSubscribe) {
-                this._activatePortDevice(port.value, type, this._getModeForDeviceType(type), 0x00);
+                this.subscribe(port.id, this._getModeForDeviceType(type));
```
aileo: This changed only to have the exact same result using manual subscribe or autosubscribe.

src/lpf2hub.ts
```diff
@@ -294,8 +294,8 @@ export class LPF2Hub extends Hub {
             const hwVersion = LPF2Hub.decodeVersion(data.readInt32LE(7));
             const swVersion = LPF2Hub.decodeVersion(data.readInt32LE(11));
             modeInfoDebug(`Port ${toHex(data[3])}, hardware version ${hwVersion}, software version ${swVersion}`);
-            this._sendPortInformationRequest(data[3]);
         }
+        this._sendPortInformationRequest(data[3]);
```
aileo: Ask for port information even if no debug flag to get parsing format

src/lpf2hub.ts
```diff
@@ -439,174 +453,117 @@ export class LPF2Hub extends Hub {
...
-                     * @param {string} port
-                     * @param {number} speed
-                     */
-                    const speed = data.readInt16LE(4);
-                    this.emit("speed", port.id, speed);
-                    break;
+    private _emitSensorEvent(port: Port, mode: IPortMode, values: number[]) {
+        switch (mode.name) {
+            case "COLOR": {
+                this.emit("color", port.id, values[0]);
```
aileo: Over all the breaking changes, it reads colors as a signed int so "no color" is -1. I think it make more sense but this is a very personal point of view.

nathankellenicki:
I love this - it's something I'd hoped to try experimenting with but again you guys beat me to it. :)

One thing I notice though is that the ginormous switch statement with emits looks very repetitive. One solution might be to have a mapping of modes to human-friendly emit types, and use the ES6 spread operator to spread out the `values` array into the different emit arguments, based on the reported length by the mode information? Do you think that's feasible?

aileo:
> One thing I notice though is that the ginormous switch statement with emits looks very repetitive. One solution might be to have a mapping of modes to human-friendly emit types, and use the ES6 spread operator to spread out the `values` array into the different emit arguments, based on the reported length by the mode information? Do you think that's feasible?

I tried to do this first but there is some special cases where it does not work very well :
- `POS` mode exists for both motors and control+ gyro but it is obviously not the same event
- `SPEC_1` mode of boost tilt sensor emits multiple events

also, it miss the event documentation that should take place there. I will look for a good compromise.

aileo:
I tried something to remove the event big switch. It hits me that with the #49 and specific class per device the edge cases would disappear.

I still don't understand how to read modes sequentially. I think `Hub._registerDeviceAttachment` could read all modes before emitting the `attach` event, but I don't know if it is possible to wait for a specific sequence of messages...

nutki:
Nicely done @aileo. I was also thinking about reworking the sensor reading code. I had some similar ideas, but overall I would try to go more radical, but I am not sure how viable would that be. Specifically I would try to make `subscribe` unnecessary. Simply by registering a listener of an event with `on` the library should be able to switch to a certain mode automatically.
As for the issue of port information reporting delay, my main concern would be that the library would become very fragile to potential message drop. What if a port mode name message is missed for some internal device? Even in positive outcome the operation would be significantly delayed. Reading all port mode infos for just the internal devices on the Boost hub takes almost 2s.
I have some clue about how the official Lego SDK handles it since it leaves some files during the operation. For example a file `46-0.0.00.1000-0.0.00.1000` with contents:
```
CAPABILITIES:15
MODE_COUNT:6
INPUT_MODES:30
OUTPUT_MODES:31
ALLOWED_COMBINATIONS:DgA=
0/NAME:POWER
0/MIN_RAW:-100
0/MAX_RAW:100
0/MIN_PCT:-100
0/MAX_PCT:100
0/MIN_SI:-100
0/MAX_SI:100
0/SYMBOL:PCT
0/MAPPING_INPUT:16
0/MAPPING_OUTPUT:0
0/VALUE_FORMAT_COUNT:1
0/VALUE_FORMAT_DECIMALS:0
0/VALUE_FORMAT_FIGURES:1
0/VALUE_FORMAT_TYPE:0
1/NAME:SPEED
1/MIN_RAW:-100
1/MAX_RAW:100
1/MIN_PCT:-100
...
```
This is obviously port mode information and the file name contains the device type number and its software and hardware version. This suggests that Lego's official SDK:
* uses this information is some way
* the information can only change when the software or hardware version changes for the device (they are both present in the device attach message)

src/lpf2hub.ts
```diff
@@ -439,174 +453,182 @@ export class LPF2Hub extends Hub {
...
+            this.emit("sensor", port.id, mode, values);
+
+            this._emitSensorEvent(port, mode, values);
         }
 
     }
 
-
+    private _emitSensorEvent(port: Port, mode: IPortMode, values: number[]) {
+        const modeToEvent: { [key: string]: string } = {
```
nutki: Good idea with the event name map, but I think it would be cool it the users could also subscribe to ports using these names. Otherwise you need to remember that to fire `distance` events you need to use `PROX` or `LPF2-DETECT` depending on the device used.

src/lpf2hub.ts
```diff
@@ -439,174 +453,182 @@ export class LPF2Hub extends Hub {
...
+             * @event LPF2Hub#colorAndDistance
+             * @param {string} port
+             * @param {Color} color Uses LWP3 color codes (0 to 10, -1 is no color).
+             * @param {number} distance Distance indicator, from 10 to 0.
+             */
+            this.emit("colorAndDistance", port.id, values[0], values[1]);
+        } else if (mode.name === "POS" && port.type === Consts.DeviceType.CONTROL_PLUS_TILT) {
+            this.emit("angle", port.id, ...values);
+        } else if (modeToEvent[mode.name]) {
+            this.emit(modeToEvent[mode.name], port.id, ...values);
```
nutki: This should ideally also normalize the values according to min/max raw and si unit. For example accel in boost hub 1G = 65 while in C+ hub 1G=8000.

src/lpf2hub.ts
```diff
@@ -362,31 +362,45 @@ export class LPF2Hub extends Hub {
...
+        const portHex = toHex(data[3]);
         const mode = data[4];
         const type = data[5];
+        const port = this._getPortForPortNumber(data[3]) as Port;
+
+        if (!port) {
+            return false;
+        }
+
+        port.modes[mode] = port.modes[mode] || {};
```
nutki: `port.modes` should be reset on device detach event. Otherwise the intoformation for a removable port will be a mix of old and new device data.

nathankellenicki:
> I tried something to remove the event big switch. It hits me that with the #49 and specific class per device the edge cases would disappear.

Agreed. I'm about to push my (very) WIP branch with these changes so you can take a look at the approach, but I think it should allow for per device mode types.

> I still don't understand how to read modes sequentially. I think `Hub._registerDeviceAttachment` could read all modes before emitting the `attach` event, but I don't know if it is possible to wait for a specific sequence of messages...

I think some kind of counter would be required, ie. device reports 4 modes (totalModes), so increment a counter (reportedModes) whenever a piece of mode info comes in. When reportedModes => totalModes, resolve a promise so that higher order code can continue.

nathankellenicki:
> Nicely done @aileo. I was also thinking about reworking the sensor reading code. I had some similar ideas, but overall I would try to go more radical, but I am not sure how viable would that be. Specifically I would try to make `subscribe` unnecessary. Simply by registering a listener of an event with `on` the library should be able to switch to a certain mode automatically.

This is fantastic idea, I hadn't thought about that! I'm literally going to start implement this in my working branch for https://github.com/nathankellenicki/node-poweredup/issues/49 - I was about to work in subscribe methods, but this is a step up.

The question of combined ports becomes an issue - ie. If you attach an event handler for "color", and one for "reflectivity", it should probably set up combined mode for the device so both can be reported. But this can be tackled later (using @aileo's code here in this PR as a base) - v1 can be one at a time.

> As for the issue of port information reporting delay, my main concern would be that the library would become very fragile to potential message drop.

Can BLE drop messages? I thought it was a "reliable" transport protocol like TCP, however I understand that some devices drop mesages in low memory situations. However how common will this be with PUP devices?

> This is obviously port mode information and the file name contains the device type number and its software and hardware version. This suggests that Lego's official SDK:
> 
> * uses this information is some way
> * the information can only change when the software or hardware version changes for the device (they are both present in the device attach message)

This is actually a great idea (essentially caching the information for known firmware versions). We could probably follow this approach?

src/hub.ts
```diff
@@ -161,12 +161,30 @@ export class Hub extends EventEmitter {
...
                 newMode = mode;
             }
+            this._ports[port].mode = newMode;
             this._activatePortDevice(this._portLookup(port).value, this._portLookup(port).type, newMode, 0x00, () => {
                 return resolve();
             });
         });
     }
 
+    public subscribeByModeName (port: string, mode: string = "") {
```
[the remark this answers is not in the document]
aileo: It definitly should call subscribe.

src/hub.ts
```diff
@@ -161,12 +161,30 @@ export class Hub extends EventEmitter {
...
+        const { value, type, modes } = this._portLookup(port);
+
+        return new Promise((resolve, reject) => {
+            this._activatePortDevice(
+                value,
+                type,
+                modes.reduce(
+                    (newMode, definition, index) => definition.name === mode ? index : newMode,
+                    this._getModeForDeviceType(type)
```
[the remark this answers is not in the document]
aileo: Yes, you are right, the current state is not reliable at all...

src/lpf2hub.ts
```diff
@@ -439,174 +453,182 @@ export class LPF2Hub extends Hub {
...
-                     * @param {number} speed
-                     */
-                    const speed = data.readInt16LE(4);
-                    this.emit("speed", port.id, speed);
-                    break;
+        if (port && port.connected && port.mode && port.modes[port.mode]) {
+            const mode = port.modes[port.mode];
+            const values = [];
+
+            for (let index = 4; index < data.length; index += Consts.VALUE_SIZE[mode.valueType]) {
```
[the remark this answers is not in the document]
aileo: Port mode informations refer to signed values, I think they could add more types in the future and we will have to follow. It makes me think that it should have some "unsupported value type" to prevent misreading.

src/lpf2hub.ts
aileo: There is aleady `this.emit("sensor", port.id, mode, values);` and it would make a lot of duplicated events, I don't know what approach should be adopted.

src/lpf2hub.ts
```diff
@@ -439,174 +453,182 @@ export class LPF2Hub extends Hub {
...
+            /**
+             * Emits when an accelerometer sensor is activated.
+             * @event LPF2Hub#angle
+             * @param {string} port
+             * @param {number} x
+             * @param {number} y
+             * @param {number} z
+             */
+            ACCEL: "acceleration",
+            ROT: "acceleration"
```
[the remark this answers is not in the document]
aileo: I totaly missed those, I did not see GRV mode.

aileo:
Thanks @nutki for the review, you always find all my mistakes, and they are numerous.

> > Nicely done @aileo. I was also thinking about reworking the sensor reading code. I had some similar ideas, but overall I would try to go more radical, but I am not sure how viable would that be. Specifically I would try to make `subscribe` unnecessary. Simply by registering a listener of an event with `on` the library should be able to switch to a certain mode automatically.
> 
> This is fantastic idea, I hadn't thought about that! I'm literally going to start implement this in my working branch for #49 - I was about to work in subscribe methods, but this is a step up.
> 
> The question of combined ports becomes an issue - ie. If you attach an event handler for "color", and one for "reflectivity", it should probably set up combined mode for the device so both can be reported. But this can be tackled later (using @aileo's code here in this PR as a base) - v1 can be one at a time.

I think it could be great as it makes events/mode relation very simple for the user. My concerns are over the complexity it adds to the lib:
- As all modes are not compatible with others (even with combination), we have to manage `off` and it forces developers to keep a reference to the listener. `removeAllListener` can be the solution for lazy  developers (me).
- If the lib override `on` and `off`, it should also override `addListener`, `once`, `removeListener`, `removeAllListener`, `prependListener` and `prependOnceListener`. This could eventually be handled by using `newListener` and `removeListener` event to make subscribe/unsubscribe logic.

src/lpf2hub.ts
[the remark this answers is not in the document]
nutki: I couldn't find any official info about those values being signed or unsigned. [This](https://lego.github.io/lego-ble-wireless-protocol-docs/index.html#value-format) does not mention it and [this](https://lego.github.io/lego-ble-wireless-protocol-docs/index.html#port-value-single) specifically refers to unsigned, which is not true in many cases.
From the empirical point of view there is a number of 8 bit port modes with raw max 255 and 16 bit with max 65535. Mostly unused modes, most relevant of them being LED RGB values (0-255), but this is only output mode.

src/lpf2hub.ts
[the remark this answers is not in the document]
nutki: It is on the previously not supported port 0x61 mode 0, I added it in a recently merged PR.

src/lpf2hub.ts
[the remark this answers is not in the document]
aileo: I can't retreive from where I understood it was signed so I must be wrong. I think we could rely on the min value and if it is under 0 then it is signed.

src/lpf2hub.ts
aileo: Do you have any link to the device/firmware mode information files ? I would try to work on a normalization helper to convert values.

src/lpf2hub.ts
aileo: this was totaly wrong

nathankellenicki:
> I think it could be great as it makes events/mode relation very simple for the user. My concerns are over the complexity it adds to the lib:
> 
> * As all modes are not compatible with others (even with combination), we have to manage `off` and it forces developers to keep a reference to the listener. `removeAllListener` can be the solution for lazy  developers (me).
> * If the lib override `on` and `off`, it should also override `addListener`, `once`, `removeListener`, `removeAllListener`, `prependListener` and `prependOnceListener`. This could eventually be handled by using `newListener` and `removeListener` event to make subscribe/unsubscribe logic.

@aileo I think you presented perfectly good solutions to those concerns. `removeAllListeners` can be used to remove any listeners the user has added, and the `newListener` can be used to hook into when new events are added (and `removeListener` when removed).

In the event the user isn't properly removing listeners, and until combined mode is added, the "last added wins" approach can work, as it's an existing paradigm that makes sense to users (up to now in this lib, you could only set one mode at a time).

@nutki I've actually implemented your suggestion in https://github.com/nathankellenicki/node-poweredup/pull/61 and I think it's a great solution. Currently I'm not working on combined modes (this is MVP1) and still need to implement unsubscribe, but it's working great.

nathankellenicki:
@aileo I'd actually love to know your thoughts on how to merge some of this with https://github.com/nathankellenicki/node-poweredup/pull/61.

I've managed to eliminate the "Big Ugly Switch Statement(TM)" of mode numbers to event names by creating a map of them and having it handled automatically, but for parsing of the sensor messages, I still have a switch statement.

I think maybe some of this could work with that, by reverse looking up the event name from the currently set mode?

aileo:
> @aileo I'd actually love to know your thoughts on how to merge some of this with #61 

I think we have to drop the part of this PR where it get parsing information from mode info as it lead to big latency (and it totally rely on not missing any message). Some sort of configuration could do the job (I don't think mode are updated that often and we could maintain some firmware related configuration like @nutki showed [here](https://github.com/nathankellenicki/node-poweredup/pull/60#issuecomment-562728382) if needed.

From what I investigated about modes in #56 , modes do not have the same position (number, id) on every hub (L/XL motors have load mode on PUP and C+ at 0x04 and CALIB at 0x05 but CALIB is at 0x04 and STATS at 0x05 on Boost).

I think we have a list of modes for each device type with :
- a name
- information about their parsing
  - value type (8/16/32 bits, floats)
  - number of values
  - min/max
  - unit
  - all the fancy stuff we can imagine (conversion function, coefficient, ...)
- their number/id depending on the hub type

We can map event name to mode name or add directly event name to the modes configuration.

I would make this configuration a static attribute of each device class instead of passing it to `super`. Maybe it could be a json file so we can track changes easily and generate it with some npm script using this lib (something like #51) ?

I would also defer the parsing to the device class when it comes to device sensor messages (or other pure device messages). This way the edge cases would be handle by the device itself and this could simplify some test suite implementation :) .

The first step is to collect all modes information for all devices on all hubs. I will do some as soon as I get my hands back on my hubs.

Side note : talking about "Big Ugly Switch Statement(TM)", a device type to constructor map would also remove the one in `hub._createDevice`

aileo:
#97 is way better than this

PR #61, opened 2019-12-07
New paradigm for handling attached devices

nathankellenicki:
This is a work in progress PR to implement some of the changes discussed in https://github.com/nathankellenicki/node-poweredup/issues/49.

Currently, large swathes of code are commented out. The approach I'm taking for this is a bottom up approach - starting with minimal functionality and adding upon it, until feature parity is nearly reached.

Currently implemented:

* Attach/detach events
* BasicMotor setSpeed and brake (to demonstrate devices sending data to the hub)
* Basic hierarchy of Device/BasicMotor/ControlPlusLargeMotor etc.

In addition, this starts to implement an auto-subscribe mechanism (originally suggested by @nutki in  https://github.com/nathankellenicki/node-poweredup/pull/60). Currently it is a Big Ugly Switch Statement(TM), however my hope is that this can be reduced and automated with some of the logic implemented by @aileo in https://github.com/nathankellenicki/node-poweredup/pull/60.

I've added a temporary example at `examples/new_device_test.js` to demonstrate how the new paradigm would work.


src/colordistancesensor.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export class ColorDistanceSensor extends Device {
+
+    constructor (hub: Hub, portId: number) {
+        super(hub, portId, Consts.DeviceType.COLOR_DISTANCE_SENSOR);
+
+        this.on("newListener", (event) => {
+            switch (event) {
+                case "color":
+                    this.subscribe(0x00);
```
aileo: I wonder if we should avoid using hex notation as all const are in decimal and I see modes as almost const.

nutki: In this case I don't mind either way. But for `Consts.DeviceType` I would actually prefer to use hex literals as this is how they are listed in [LWP doc](https://lego.github.io/lego-ble-wireless-protocol-docs/index.html#io-type-id).

src/basicmotor.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+
+    /**
+     * Set the motor speed.
+     * @method BasicMotor#setSpeed
+     * @param {number} speed For forward, a value between 1 - 100 should be set. For reverse, a value between -1 to -100. Stop is 0.
+     * @returns {Promise} Resolved upon successful completion of command.
+     */
+    public setSpeed (speed: number) {
```
nutki: Since basic motors don't have a tachometer, this is actually `setPower`. I guess this makes it easier for the API, but would you then have `setSpeed` and `setPower` for tacho motors or just `setSpeed`?

src/device.ts
```diff
@@ -0,0 +1,56 @@
...
+    private _busy: boolean = false;
+    private _finished: (() => void) | null = null;
+
+    constructor (hub: Hub, portId: number, type: number = Consts.DeviceType.UNKNOWN) {
+        super();
+        this._hub = hub;
+        this._portId = portId;
+        this._type = type;
+        this.hub.on("detach", (device) => {
+            if (device.portId === this.portId) {
```
nutki: Wouldn't `device === this` be better? This way old disconnected devices would not generate `detach` events.

src/device.ts
```diff
@@ -0,0 +1,56 @@
...
+    public get port () {
+        return this.hub.getPortNameForPortId(this.portId);
+    }
+
+    public get type () {
+        return this._type;
+    }
+
+    public send (data: Buffer, characteristic: string = Consts.BLECharacteristic.LPF2_ALL, callback?: () => void) {
+        this.hub.send(data, characteristic, callback);
```
nutki: Doesn't this send `LPF2_ALL` characteristic messages to WeDo2 hub? As a side note, I am not sure if continued support of the WeDo hub makes sense. The new devices will not be supported on it anyway since the FW is not updatable (according to a Lego designer) and it has limited availability (which also makes it expensive).

src/hub.ts
```diff
@@ -18,10 +19,12 @@ export class Hub extends EventEmitter {
 
     public autoSubscribe: boolean = true;
     public useSpeedMap: boolean = true;
-    public type: Consts.HubType = Consts.HubType.UNKNOWN;
+    protected _type: Consts.HubType = Consts.HubType.UNKNOWN;
 
-    protected _ports: {[port: string]: Port} = {};
-    protected _virtualPorts: {[port: string]: Port} = {};
+    protected _attachedDevices: Device[] = [];
```
nutki: Given the current use I would say a map of portId to Device would make more sense than a flat array for `_attachedDevices`.

src/colordistancesensor.ts
nathankellenicki: It's a good question. I think for the moment having the values hardcoded is a placeholder, I'd like to put all consts in a file somewhere. So - hex or decimal? Perhaps hex as @nutki suggests?

src/basicmotor.ts
nathankellenicki: Good point. I guess I haven't liked `power` as I think for most users `speed` is a more appropriate term for what they want, even if it isn't exactly the same. I don't this library should attempt to exactly emulate LWP functionally, instead adding a user friendly layer on top of it.

Perhaps we could have `setSpeed(50)` set power by default, however with an options object (such as `setSpeed(50, { maxPower: 50 })` to switch to maintaining the speed with power?

src/device.ts
nathankellenicki: Good catch. Will change this.

src/device.ts
nathankellenicki: Right now, yes, eventually I want to have each device method (ie. `setSpeed`) switch characteristics and message based on hub type.

Regarding the WeDo 2.0 support - I've also been thinking about this recently, but I think I want to keep support personally, as I actually use a few WeDo 2.0 hubs myself. :) They are currently the only hub type that offers an AC adapter so it can run forever.

On the topic of firmware support, that's correct - however most newer devices can be supported on it in some way. See the compatibility matrix in the readme of this project.

Eventually I might drop it though - especially if Lego releases a hub with an AC adapter.

src/colordistancesensor.ts
aileo: Yes, like @nutki wrote hex make more sense regarding the doc. Even more if we had some const with all message headers and all.

Plus, the end users can use whatever format they want when they call methods.

src/hub.ts
nathankellenicki: I think so too. But object literals don't accept numbers as key type, and I was hoping not to use ES6 maps. I guess we could though (_"Just shut up and do it, Nathan!"_)

nathankellenicki: I'm totally wrong - they *do* support numbers as keys/property names, just not in initializers. I'll change to use this. :)

src/device.ts
nutki: Maybe a more straightforward approach would be to remove this listener on detach.

src/device.ts
nutki: Yeah, I understand that WeDo support is mostly for people who already own one. Iteresting about the AC adapter though, I have not heard about it. Is this stand alone or the one with rechargeable battery allows to bypass the battery? Long time ago I especially for that purpose bought Lego [RCX with an AC plug](https://www.bricklink.com/v2/catalog/catalogitem.page?P=884a#T=I), even though I already had 2 without :)

src/hub.ts
nutki: Do you mean initializes as in: `({0:'Test'})[0]`? It does work for me. What's wrong with ES6 Maps, btw? I strongly prefer them to using POJOs for maps, especially in Typescript.

src/pupremote.ts
```diff
@@ -37,12 +37,12 @@ export class PUPRemote extends LPF2Hub {
...
+    constructor (device: IBLEAbstraction, autoSubscribe: boolean = true) {
         super(device, autoSubscribe);
-        this.type = Consts.HubType.POWERED_UP_REMOTE;
-        this._ports = {
-            "LEFT": new Port("LEFT", 0),
-            "RIGHT": new Port("RIGHT", 1)
+        this._type = Consts.HubType.POWERED_UP_REMOTE;
+        this._portNames = {
+            "LEFT": 0,
+            "RIGHT": 1
```
aileo: This looks repetitive over hubs, does it make sense to use the same approach you have on devices ?

```js
super(device, autosubscribe, Consts.HubType.POWERED_UP_REMOTE, {
    "LEFT": 0,
    "RIGHT": 1
});
```

examples/winter_village_train.js
```diff
@@ -0,0 +1,60 @@
...
+    } else if (hub.name === "NK_Winter_Sensor") {
+        await hub.connect(); // Connect to hub
+        console.log(`Connected to sensor!`);
+        sensor = hub;
+
+        sensor.on("distance", (_, distance) => {
+            if (distance < 5 && !ramping) {
+                await stopTrain();
+            }
+        });
```
aileo: This listen a device (sensor) event on the hub, does it mean we support device events at the hub level ?

nathankellenicki:
Oops, I shouldn’t have committed that! :) It’s for my house - designed to run on the old code (master).

nathankellenicki:
Yeah, perhaps so. To be honest I’ve barely touched the hub setup, I’ve been focused on the device abstraction. But they will get a revamp too.

examples/winter_village_train.js
nathankellenicki: Oops, I shouldn’t have committed that! :) It’s for my house - designed to run on the old code (master). But I haven’t though more about supporting that as of now.

src/pupremote.ts
nathankellenicki: Yeah, perhaps so. To be honest I’ve barely touched the hub setup, I’ve been focused on the device abstraction. But they will get a revamp too.

examples/winter_village_train.js
aileo: It makes a nice example !

src/device.ts
nathankellenicki: > Yeah, I understand that WeDo support is mostly for people who already own one. Iteresting about the AC adapter though, I have not heard about it. Is this stand alone or the one with rechargeable battery allows to bypass the battery? Long time ago I especially for that purpose bought Lego [RCX with an AC plug](https://www.bricklink.com/v2/catalog/catalogitem.page?P=884a#T=I), even though I already had 2 without :)

It's the rechargeable battery pack for the WeDo 2.0 hub - it has a built in port to plug in an AC adapter. :)

https://education.lego.com/en-us/products/wedo-2-0-add-on-power-pack-by-lego-education/5004838

That will allow you to keep a WeDo 2.0 hub running forever.

src/boostmovehub.ts
```diff
@@ -34,19 +34,15 @@ export class BoostMoveHub extends LPF2Hub {
...
-            "D": new Port("D", 3),
-            "TILT": new Port("TILT", 58)
+    constructor (device: IBLEAbstraction) {
+        super(device, Consts.HubType.BOOST_MOVE_HUB);
+        this._portNames = {
+            "A": 0,
+            "B": 1,
+            "C": 2,
+            "D": 3,
+            "TILT": 58
```
aileo: I think port names/id should appear in consts as they are constant and the same over same type hubs. It would also help developers like @ractive and me determine which port we should look at in our projects.

src/colordistancesensor.ts
nathankellenicki: @aileo I've started getting rid of the hex notation and implementing modes as enums. And as suggested by @nutki somewhere else, subscriptions are automatically made when attaching the associated event handler. It's much cleaner, thanks!

src/basicmotor.ts
nathankellenicki: @nutki I've gone with `setPower` and done a 180 on this. :) Will implement `setSpeed` appropriately later.

src/device.ts
nathankellenicki: @nutki Listener is now removed on detach.

src/boostmovehub.ts
nathankellenicki: I've now created a const (exported under the hub namespace) for each hub, accessible via `PoweredUP.WeDo2SmartHub.PortMap`, etc. This is also passed to the root hub object at constructor time. It's also available at runtime via a `ports` getter, ie. `hub.ports = ["A", "B", ...]`.

src/pupremote.ts
nathankellenicki: This is also now done - https://github.com/nathankellenicki/node-poweredup/pull/61#discussion_r359069774

src/device.ts
nathankellenicki: I've implemented some WeDo2 support into some devices - it's actually on par with master feature wise, but I may implement some more modes, once I test to see what works and doesn't!

It's much simpler to implement WeDo2 support with each motor/sensor having its own device class.

examples/winter_village_train.js
aileo: the new `waitForDeviceByType` should allow you to migrate this to the new paradigm

nathankellenicki: Yeah - it was an idea that only occurred to me last night while thinking about other stuff, but it solves a problem that many people have, ie. not knowing when to call `getDeviceByType` without watching for an `attach` event, by which point it's provided to you anyway. It just makes it easier to maintain a more sequential coding pattern, which is easier for a lot of people.

nathankellenicki: It's also how I was this could support internal motors/functions on a hub rather on device. ie. If a boost internal motor has a method `setSpeed(speed)`, I can create a lightweight wrapper on the hub called `setSpeed(port, speed)`, which simply passes through the values after using `waitForDeviceByPort`.

aileo: Awesome

src/interfaces.ts
```diff
@@ -13,3 +15,10 @@ export interface IBLEDevice extends EventEmitter {
     readFromCharacteristic: (uuid: string, callback: (err: string | null, data: Buffer | null) => void) => void;
     writeToCharacteristic: (uuid: string, data: Buffer, callback?: () => void) => void;
 }
+
+export interface IDeviceInterface extends EventEmitter {
+    type: Consts.HubType;
+    getPortNameForPortId: (portId: number) => string | undefined;
+    send: (message: Buffer, uuid: string, callback?: () => void) => void;
+    subscribe: (portId: number, deviceType: number, mode: number) => void;
+}
```
aileo: Even if passed to devices, it seems to be a Hub interface.

nathankellenicki: Yeah - I've had to do this for the moment as TypeScript doesn't like circular dependencies/definitions - Hub imports Device which imports Hub (even if for the type only). This was a quick fix - I plan to revisit this.

aileo: My concern was more about the naming, I took it for a device definition at first.

src/devices/hubled.ts
```diff
@@ -45,28 +45,7 @@ export class HubLED extends Device {
...
-    public setBrightness (brightness: number) {
-        return new Promise((resolve) => {
-            if (this.isWeDo2SmartHub) {
-                const data = Buffer.from([this.portId, 0x01, 0x02, brightness]);
-                this.send(data, Consts.BLECharacteristic.WEDO2_MOTOR_VALUE_WRITE);
-            } else {
-                const data = Buffer.from([0x81, this.portId, 0x11, 0x51, 0x00, brightness]);
-                this.send(data);
-            }
+            this.writeDirect(0x00, Buffer.from([red, green, blue]));
```
aileo: Regarding the diff, it seems it should be `this.writeDirect(0x01, Buffer.from([red, green, blue]));`.

src/devices/device.ts
```diff
@@ -77,6 +77,14 @@ export class Device extends EventEmitter {
         return this._isWeDo2SmartHub;
     }
 
+    public writeDirect (mode: number, data: Buffer, callback?: () => void) {
+        if (this.isWeDo2SmartHub) {
+            this.send(Buffer.concat([Buffer.from([this.portId, 0x01, 0x02]), data]), Consts.BLECharacteristic.WEDO2_MOTOR_VALUE_WRITE);
+        } else {
+            this.send(Buffer.concat([Buffer.from([0x81, this.portId, 0x11, 0x51, mode]), data]), Consts.BLECharacteristic.LPF2_ALL, callback);
```
aileo: Does it change the subscribed mode ?

nathankellenicki: I'm not exactly sure. I don't think it changes subscribed mode _per se_. I think this changes the output mode of the device, ie. switching between accepting a triple RGB value vs a single color value, but doesn't actually subscribe to any notifications. But I have noticed in some scenarios it's necessary to send an explict subscribe first.

src/devices/hubled.ts
nathankellenicki: Well spotted!

src/devices/device.ts
aileo: I noticed two output modes on the color and distance : COL_O & IR_Tx.

I assume COL_O is some kind of color control for the led, I don't have one right now but we could  be able to understand what it does from its mode definition.

I think IR_Tx is some kind of serial message sending over IR, I don't have any use case for this but I find this idea somehow very exiting.

I wonder if having a mode definition is the difference between needing to subscribe before or not.

src/hubs/basehub.ts
```diff
@@ -48,10 +54,11 @@ export class BaseHub extends EventEmitter {
 
     private _type: Consts.HubType;
     private _portMap: {[portName: string]: number} = {};
-    private _attachCallbacks: ((device: Device) => void)[] = [];
+    private _attachCallbacks: ((device: Device) => boolean)[] = [];
 
     constructor (device: IBLEAbstraction, portMap: {[portName: string]: number} = {}, type: Consts.HubType = Consts.HubType.UNKNOWN) {
         super();
+        this.setMaxListeners(20); // Technic Medium Hub has 9 built in devices + 4 external ports. Node.js throws a warning after 11 attached event listeners.
```
aileo: You can also add a listener for each device type by using `waitForDeviceByType`. There is 23 types + the 13 ports on the Technic Medium Hub and we forgot about the 6 possible virtuals ports.

Lets say somebody is crazy enougth to listen to everything, you could have up to 42 listeners.

nathankellenicki: Sorry my comment is shorthand - the eventListener limit is _per event type_ - in particular the warning is being thrown here is because when the hub detects a new attached device, it creates a new device object, which in turn listens to the "detach" event on the hub to know when to consider itself detached.

I've tested that they're successfully removed, so on a Technic Medium Hub there should never be more than 13 detach events added by this library, plus any extra the user wants (hopefully a single detach event!).

nathankellenicki: Ps. `waitForDeviceByType` doesn't use any event listeners - it merely stashes a callback to be called on an attach, to resolve a promise.

aileo: I totally brainfarted. Proof I should stop reviewing past midnight...

nathankellenicki: No worries 😄  Thanks for the diligence, it's appreciated! (and necessary!)

src/hubs/basehub.ts
```diff
@@ -0,0 +1,407 @@
...
+                break;
+            case Consts.DeviceType.DUPLO_TRAIN_BASE_SPEEDOMETER:
+                device = new DuploTrainBaseSpeedometer(this, portId);
+                break;
+            default:
+                device = new Device(this, portId, undefined, deviceType);
+                break;
+        }
+
+        return device;
```
aileo: Get rid of the switch :

```js
const deviceConstructors: {[type in Consts.DeviceType]: typeof Device} = {
            [Consts.DeviceType.LIGHT]: Light,
            [Consts.DeviceType.TRAIN_MOTOR]: TrainMotor,
            [Consts.DeviceType.SIMPLE_MEDIUM_LINEAR_MOTOR]: SimpleMediumLinearMotor,
            [Consts.DeviceType.MOVE_HUB_MEDIUM_LINEAR_MOTOR]: MoveHubMediumLinearMotor,
            [Consts.DeviceType.MOTION_SENSOR]: MotionSensor,
            [Consts.DeviceType.TILT_SENSOR]: TiltSensor,
            [Consts.DeviceType.MOVE_HUB_TILT_SENSOR]: MoveHubTiltSensor,
            [Consts.DeviceType.TECHNIC_MEDIUM_HUB_TILT_SENSOR]: TechnicMediumHubTiltSensor,
            [Consts.DeviceType.TECHNIC_MEDIUM_HUB_GYRO_SENSOR]: TechnicMediumHubGyroSensor,
            [Consts.DeviceType.TECHNIC_MEDIUM_HUB_ACCELEROMETER]: TechnicMediumHubAccelerometerSensor,
            [Consts.DeviceType.MEDIUM_LINEAR_MOTOR]: MediumLinearMotor,
            [Consts.DeviceType.TECHNIC_LARGE_LINEAR_MOTOR]: TechnicLargeLinearMotor,
            [Consts.DeviceType.TECHNIC_XLARGE_LINEAR_MOTOR]: TechnicXLargeLinearMotor,
            [Consts.DeviceType.COLOR_DISTANCE_SENSOR]: ColorDistanceSensor,
            [Consts.DeviceType.VOLTAGE_SENSOR]: VoltageSensor,
            [Consts.DeviceType.CURRENT_SENSOR]: CurrentSensor,
            [Consts.DeviceType.REMOTE_CONTROL_BUTTON]: RemoteControlButton,
            [Consts.DeviceType.HUB_LED]: HubLED,
            [Consts.DeviceType.DUPLO_TRAIN_BASE_COLOR_SENSOR]: DuploTrainBaseColorSensor,
            [Consts.DeviceType.DUPLO_TRAIN_BASE_MOTOR]: DuploTrainBaseMotor,
            [Consts.DeviceType.DUPLO_TRAIN_BASE_SPEAKER]: DuploTrainBaseSpeaker,
            [Consts.DeviceType.DUPLO_TRAIN_BASE_SPEEDOMETER]: DuploTrainBaseSpeedometer
        };

        return new deviceConstructors[deviceType](this, portId);
```

nathankellenicki: It's a good shout, will look into doing something like this. I actually want to get rid of all the global consts and have them all under their relevant classes, but it should work the same regardless.

aileo: > [...] I actually want to get rid of all the global consts and have them all under their relevant classes[...]

I tried something like this in #65 with 64425a9a7df3331229592076a07b042f13a9dca8

src/hubs/basehub.ts
```diff
@@ -0,0 +1,369 @@
...
+    private _portMap: {[portName: string]: number} = {};
+    private _attachCallbacks: ((device: Device) => boolean)[] = [];
+
+    constructor (device: IBLEAbstraction, portMap: {[portName: string]: number} = {}, type: Consts.HubType = Consts.HubType.UNKNOWN) {
+        super();
+        this.setMaxListeners(20); // Technic Medium Hub has 9 built in devices + 4 external ports. Node.js throws a warning after 11 attached event listeners.
+        this._type = type;
+        this._bleDevice = device;
+        this._portMap = portMap;
+        device.on("disconnect", () => {
```
aileo: I think it should be named `bleDevice` as for the attribute because `device.on("disconnect", () => {})` can have a very diferent meaning with the new paradigm

src/interfaces.ts
nathankellenicki: Makes sense - will change

src/utils.ts
```diff
@@ -38,3 +38,20 @@ export const normalizeAngle = (angle: number) => {
...
+    if (angle < -45) {
+        return -90;
+    }
+    if (angle < 45) {
+        return 0;
+    }
+    if (angle < 135) {
+        return 90;
+    }
+    return -180;
```
aileo: I would go for a more mathematical approach :
```js
export const roundAngleToNearest90 = (angle: number) => {
    return normalizeAngle(Math.round(angle/90)*90);
}
```

nathankellenicki:
I think at this point this PR has reached feature parity with what was there before. I'd love some help testing it before I end up approving it and merging it to master and pushing out a new release.

I'm hoping to do this soon - especially as my copy of Spike Prime is scheduled to be delivered today, so I'll have some new devices to add support for shortly! :)

aileo:
> I'd love some help testing it before I end up approving it and merging it to master and pushing out a new release.

Do you have any testing plan ? I have no idea where to start.

Aslo, I can only test the web-bluetooth part with my current setup.

nathankellenicki:
> > I'd love some help testing it before I end up approving it and merging it to master and pushing out a new release.
> 
> Do you have any testing plan ? I have no idea where to start.
> 
> Aslo, I can only test the web-bluetooth part with my current setup.

My testing plan up until now has been quite simple - for every hub, try every sensor in every mode, and every motors output. :D It's not especially quick.

Thus far I'm happy with this PR, though more testing (both similar to the plan and real-life application testing) I'm always thankful for!

I think the main things I need to do now are update the documentation and JSDoc comments, and update the samples.

Issue #63, opened 2019-12-08
Port infos on Control+ hub not always available


ractive:
I'm implementing a react based remote control app using node-poweredup:
https://brick-remote.ractive.ch
I'm testing and developing this with the Control+ hubs.

The issue is that when a hub is connected, its port infos are not always available. 

In this example ports A, B and D are connected, but port A never shows any info, port B most of the time and port D almost always:
<img src="https://user-images.githubusercontent.com/783861/70382179-d2efc700-1957-11ea-946f-12433de080a5.png" height="60%" width="60%" />

<img src="https://user-images.githubusercontent.com/783861/70382202-6aedb080-1958-11ea-8257-7dbf99bf7bdf.png" height="60%" width="60%" />

I'm getting the port info with `hub.getPortDeviceType(port)` (see [here](https://github.com/ractive/brick-remote/blob/300ce2aafc84523f9b702ec03a6d7522e06a5a9e/src/components/HubDetails.tsx#L111)) after a call to `await hub.connect()` (see [here](https://github.com/ractive/brick-remote/blob/9179b526d4271ba2e37a76e15259eb86ea131378/src/App.tsx#L123)).
The ports that are identified as "unkown" also don't get sensor messages. While e..g port B gets "rotate" events, port A does not.

Steering the motors on those ports although works and e.g. a motor on port A that is recognized as "unknown" can be successfully controlled by calling `hub.setMotorSpeed("A", 50)` (see [here](https://github.com/ractive/brick-remote/blob/2d7e57085ac815bea9aee35f3456ec20ba48a4b9/src/components/MotorControl.tsx#L69)).

Any ideas what is wrong here so that the port infos are not correctly set after the hub is connected?

The source code of the app can be found [here](https://github.com/ractive/brick-remote ) and started locally with `yarn start dev`. Debug logging is currently enabled (see [here](https://github.com/ractive/brick-remote/blob/ce8c3ef8ec96da1b43056c62351b24bb17887e5c/public/index.html#L4)).

nathankellenicki:
The reason for this is that you aren’t waiting to query for attached devices after connecting. Attached devices aren’t available immediately after connection - there is a handshake process, and then the hub discovers and advertises attached devices.

You can wait for the “attach” event to be emitted from the hub for the port you want. At that point you’ll know it’s available and ready.

ractive:
I added an event listener for the "attach" events like this:
```
[...]

PR #64, opened 2019-12-13
Characteristic change callback instead of event listener

aileo:
Related to #63, trying to not miss any message by keeping event listener and changing callback.

### DIFF
--- a/src/webbledevice.ts
+++ b/src/webbledevice.ts
@@ -85,6 +85,18 @@ export class WebBLEDevice extends EventEmitter implements IBLEDevice {
             const characteristics = await service.getCharacteristics();
             for (const characteristic of characteristics) {
                 this._characteristics[characteristic.uuid] = characteristic;
+                // @ts-ignore
+                characteristic.addEventListener("characteristicvaluechanged", (event) => {
+                    if (this._listeners[characteristic.uuid] && typeof this._listeners[characteristic.uuid] === "function") {
+                        const buf = Buffer.alloc(event.target.value.buffer.byteLength);
+                        const view = new Uint8Array(event.target.value.buffer);
+                        for (let i = 0; i < buf.length; i++) {
+                            buf[i] = view[i];
+                        }
+                        debug("Incoming data", buf);
+                        return this._listeners[characteristic.uuid](buf);
+                    }
+                });
             }
             debug("Service/characteristic discovery finished");
             return discoverResolve();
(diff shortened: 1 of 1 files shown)


nathankellenicki:
I'm unable to test this right now (I fly back home tonight) but I suspect this might cause a few problems. Normally LPF2 hubs don't start sending data until there's a subscriber to the characteristic asking for notifications, so subscribing that early may cause data to be sent before we're ready for it.

Also in the case of WeDo 2.0 hubs, there are a lot more charactistics on there that we care about, so subscribing to all of them indiscriminately could cause a lot more chatter and therefore BLE latency.

aileo:
It should not start sending as it does not call startNotification, it only add the listener.

nathankellenicki:
> It should not start sending as it does not call startNotification, it only add the listener.

@aileo I see, sorry I missed that.

I think I'm still missing something though. So now there's a persistent listener on the characteristic event, which does the conversion, and eventually calls the call back on `_listeners`. However, if there's nothing on `_listeners`, the message will get lost anyway? What problem is this solving?

Sorry if I'm misunderstanding the code/purpose. :)

aileo:
It "solve" the problem, that we suspect does not exist: removing and adding the listener.

This instead change the function called in the listener so it guarantee:
- no message can be lost due to no listener 
- a message can't be read twice as there is not two listeners at the same time (even during the same tick).

nathankellenicki:
> This instead change the function called in the listener so it guarantee:
> 
> * no message can be lost due to no listener

How does it solve this problem? There is always a listener yes, but that listener calls a callback with the message. If there is no callback, it doesn't do anything with the message, losing it. If we want to take this approach, should we put the message into the mailbox to be redelivered when a new callback is added?

It seems this is just moving the same problem a little further down the tree...

nathankellenicki:
> * a message can't be read twice as there is not two listeners at the same time (even during the same tick).

I can see it solves this edge case. We could accept this PR as a fix for this, and think some more about lost messages?

aileo:
Sorry, my intent was not clear, it is only meant to prevent message drop from the event listener change as I mentionned in [#63](https://github.com/nathankellenicki/node-poweredup/issues/63#issuecomment-565595879):

> [...] And the old event listener is removed before the new one is added but I doubt it would lead to miss a message.


> We could accept this PR as a fix for this, and think some more about lost messages?

I think we should not, it seems it could cause more problems than it solves

> should we put the message into the mailbox to be redelivered when a new callback is added

I will take this approach, do you think this part should handle the multiple frame messages ?

nathankellenicki:
> I will take this approach, do you think this part should handle the multiple frame messages ?

Thanks for the help. :) I think it's needed yep. I have seen it happen where multiple frames appear in the same payload. I was also thinking - there is the case where a single message frame can be split across multiple payloads depending on MTU, so I think the case needs to be handled where a callback gets added in the middle of a full frame arriving (ie. stashing a partial message in the mailbox and redelivered later)

Worth noting that all of these cases are handled in the more advanced message parsing inside lpf2hub.ts, but I simplified it in the WebBLE discovery - but obviously forgot to handle these edge cases.

aileo:
This is totally outdated, better start from the current version again.

PR #65, opened 2019-12-18
Feature/devices: parsing sensor

aileo:
This is not tested, just some code to illustrate [this discussion](https://github.com/nathankellenicki/node-poweredup/pull/60#issuecomment-566950937).

I only worked on the color and distance sensor so it breaks other devices. @nathankellenicki , I wanted to know if it is at your taste before going further.


src/devices/device.ts
```diff
@@ -78,16 +93,73 @@ export class Device extends EventEmitter {
...
+                }
+                case Consts.ValueType.Int32: {
+                    data.push(message.readInt32LE(index));
+                    break;
+                }
+                case Consts.ValueType.Float: {
+                    data.push(message.readFloatLE(index));
+                    break;
+                }
+            }
```
aileo: Because "We All Love Switches (TM)" :)

src/devices/device.ts
```diff
@@ -78,16 +93,73 @@ export class Device extends EventEmitter {
...
+        }
+
+        if (mode.event) {
+            this.emit(mode.event, ...data);
+        }
```
aileo: emit related event if any.

This break the distance one as it done not apply calculation on the value and also the colorAndDistance which will receive color, distance (0-10), led color and reflectivity instead of only color and distance.

src/hubs/lpf2hub.ts
```diff
@@ -410,7 +410,8 @@ export class LPF2Hub extends Hub {
         const device = this._getDeviceByPortId(portId);
 
         if (device) {
-            device.receive(message);
+            // Forward message to device (without size, hub and port)
+            device.receive(message.slice(3));
```
aileo: Device does not need to get the full message, data is sufficient.
I did not test and I wonder if it should be 4 instead of 3.

src/devices/colordistancesensor.ts
```diff
@@ -2,90 +2,111 @@ import { Device } from "./device";
...
+            input: true,
+            event: "colorAndDistance",
+            values: { type: ValueType.UInt8, count: 4, min: 0, max: 255 },
+            num: {
+                [HubType.BOOST_MOVE_HUB]: 0x08,
+                [HubType.CONTROL_PLUS_HUB]: 0x08,
+                [HubType.POWERED_UP_HUB]: 0x08
+            }
+        }
+    };
```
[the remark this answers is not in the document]
aileo: We could add here some transformation function that get data array in parameters and return transformed array. It could do some calculation for distance or simply remove some values (color and distance).

src/devices/hubled.ts
```diff
@@ -20,13 +20,10 @@ export class HubLED extends Device {
...
     public setColor (color: number | boolean) {
         return new Promise((resolve, reject) => {
-            if (this.mode !== HubLED.Mode.COLOR) {
-                this.subscribe(HubLED.Mode.COLOR);
-            }
             if (typeof color === "boolean") {
                 color = 0;
             }
-            this.send(Buffer.from([0x81, this.portId, 0x11, 0x51, 0x00, color]));
+            this.sendWithMode("COLOR", Buffer.from([0x81, this.portId, 0x11, 0x51, 0x00, color]));
```
aileo: The `0x81, this.portId, 0x11``is common to all devices sent messages. I am not sure to understand that part of the protocol yet but I think it refers to output commands.

`Device` could have some wrapper method to prepend this part and remove repetitions.

src/hubs/lpf2hub.ts
aileo: Also, it seems the offset is not the same for WEDO2, it would remove specifique parsing from devices.

src/devices/hubled.ts
```diff
@@ -75,10 +69,24 @@ export class HubLED extends Device {
...
-    }
-
+    export const modes: { [name: string]: IDeviceMode } = {
+        COLOR: {
+            input: false,
+            num: {
+                [HubType.MOVE_HUB]: 0x00,
+                [HubType.TECHNIC_MEDIUM_HUB]: 0x00,
+                [HubType.HUB]: 0x00,
+                [HubType.WEDO2_SMART_HUB]: 0x00
```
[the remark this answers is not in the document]
nathankellenicki: I do this already in a few places - for example in `currentsensor.ts` there is a default set for `HubType.UNKNOWN`. Perhaps this could follow the same model.

src/devices/hubled.ts
nathankellenicki: Agreed and good spot - this is known in the LWP3 documentation as WriteDirect mode, so perhaps I'll create a method called `writeDirect`.

nathankellenicki:
This is excellent, thank you! I have a few thoughts.

Firstly, it seems each device has an attached firmware and hardware revision. I wonder if future device production runs may have different firmware, perhaps for bug fixes - or even if device firmware can even be updated through the hub. If so, then it's conceivable that modes might change, or at least mode values (ie. min/max).

Perhaps we should do some kind of hardware/firmware version check when first connecting a device, before the device object is attached to the hub? If we have a mode definition for that version combination, we use that, otherwise we use the commands to retrieve the definition?

This would add a small amount of latency before we attach the device, but only in scenarios where we don't already have the definition embedded, and it gives us an incentive to stay up to date.

Secondly, I do like the definition aspect of this, but I think it should be kept separate from the implementation, ie. out of the device classes. Perhaps we should store them as json files to be loaded at runtime from a `definitions/` directory?

Regarding timelines for the `feature/devices` branch - you may have noticed it's been quite fast moving recently as I actively work on this. :) I hope to get an initial version out within the next few days, but I was planning to make a few small changes to devices to handle being able to attach events to the hub and not just the device. I appreciate the speediness of this PR - but perhaps it would be worth it to hold off until `feature/devices` branch is more stable, ie. on version release?

src/devices/device.ts
```diff
@@ -82,16 +97,102 @@ export class Device extends EventEmitter {
...
+            if (previousMode) {
+                this.subscribe(previousMode);
+                if (callback) {
+                    callback();
+                }
+            }
+        });
+    }
+
+    protected sendLinearPowerCommand (value: number) {
```
nathankellenicki: 👍 I like this - but not sure on the naming as this is `DirectWrite` mode and isn't just for setting power - ie. it's used for setting LED color/RGB on a hub.

src/devices/device.ts
```diff
@@ -82,16 +97,102 @@ export class Device extends EventEmitter {
...
+            // TODO : error handling -> no parsing informations
+            return;
+        }
+
+        const data = [];
+
+        for (let index = 0; index <= message.length; index += Consts.ValueBits[mode.values.type]) {
+            switch (mode.values.type) {
+                case Consts.ValueType.UInt8: {
+                    data.push(message.readUInt8(index));
```
nathankellenicki: Should the raw value be changed to take into account min/max value at this point?

src/devices/device.ts
nathankellenicki: This is the way.

src/devices/device.ts
```diff
@@ -82,16 +97,102 @@ export class Device extends EventEmitter {
...
+        });
+    }
+
+    protected sendLinearPowerCommand (value: number) {
+        return new Promise((resolve) => {
+            if (this.isWeDo2SmartHub) {
+                const data = Buffer.from([this.portId, 0x01, 0x02, value]);
+                this.send(data, Consts.BLECharacteristic.WEDO2_MOTOR_VALUE_WRITE);
+            } else {
+                const data = Buffer.from([0x81, this.portId, 0x11, 0x51, 0x00, value]);
```
nathankellenicki: There can be multiple values - the LED light takes seperate R G B values, for instance.

src/devices/hubled.ts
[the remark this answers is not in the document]
aileo: It may be weird but I don't like using `UNKNOW` as the `ALL` placeholder, it may lead to misunderstanding while reading the code.

src/devices/device.ts
aileo: I did not know it was the same header everywhere, I planned to add the mapSpeed logic directly to this method by adding min/max parameters and maybe the 127 thing.

nathankellenicki: Yeah, see here for usage when setting hub LED RGB color: https://github.com/nathankellenicki/node-poweredup/blob/feature/devices/src/devices/hubled.ts#L48

Here is the documentation for this command: https://lego.github.io/lego-ble-wireless-protocol-docs/index.html#writedirectmodedata

aileo:
> Firstly, it seems each device has an attached firmware and hardware revision. I wonder if future device production runs may have different firmware, perhaps for bug fixes - or even if device firmware can even be updated through the hub. If so, then it's conceivable that modes might change, or at least mode values (ie. min/max).

I would say it is not first priority, lets see when it happens.
Just like the mode number is defined per hub type, some part of the definition could be per firmware/hardware version.

> Perhaps we should do some kind of hardware/firmware version check when first connecting a device, before the device object is attached to the hub? If we have a mode definition for that version combination, we use that, otherwise we use the commands to retrieve the definition?
> This would add a small amount of latency before we attach the device, but only in scenarios where we don't already have the definition embedded, and it gives us an incentive to stay up to date.

This looks like a fair compromise.
 
> Secondly, I do like the definition aspect of this, but I think it should be kept separate from the implementation, ie. out of the device classes. Perhaps we should store them as json files to be loaded at runtime from a `definitions/` directory?

The definition object is not serializable since I added the `transform` method that handle special conversions.
In fact, I wonder if it is a weakness in the current model, raw min/max and SI min/max should solve this problem (at least for current/voltage sensor) if it was defined per hub type (like the mode numbers).
I tried the JSON approach at first, but loading JSON in typescript is not in my competencies. Maybe some TS namespaces files in `definitions/` directory. I wonder how to handle the "loaded at runtime" for the browser. Maybe just some filtering at "constructor" time.

> Regarding timelines for the `feature/devices` branch - you may have noticed it's been quite fast moving recently as I actively work on this. :) I hope to get an initial version out within the next few days, but I was planning to make a few small changes to devices to handle being able to attach events to the hub and not just the device. I appreciate the speediness of this PR - but perhaps it would be worth it to hold off until `feature/devices` branch is more stable, ie. on version release? But that's entirely up to you!

Absolutely no problem, I am just better at writing code than explanations :)
This is just pure theory as it does not even compile, I just wanted to see what we could do.
I don't really care about this code being thrown away as it serves as discussion support.

nathankellenicki:
Well, if we go down the JSON file route then each JSON file could be a combination of `firmware-hardware-hubtype-devicetype.json`, which would allow for making it purely definition based, and allow for version specificity with little effort, just by checking for the appropriate filename.

> I tried the JSON approach at first, but loading JSON in typescript is not in my competencies.

I'm sure we can work together on that. :)

>  I wonder how to handle the "loaded at runtime" for the browser.

Webpack can handle that automatically when bundling. It'll traverse the AST looking for import statements, and bundle the JSON files into the bundle. The `json-loader` Webpack plugin should handle it for us.

> Absolutely no problem, I am just better at writing code than explanations :)
> This is just pure theory as it does not even compile, I just wanted to see what we could do.
> I don't really care about this code being thrown away as it serves as discussion support.

Don't throw it away! I'd love to get some of this in at some point. Code definitely makes clearer what we're thinking in these contexts. :)

aileo:
> [...] it seems each device has an attached firmware and hardware revision. I wonder if future device production runs may have different firmware, perhaps for bug fixes - or even if device firmware can even be updated through the hub [...]

I am not sure, but I have the intuition that this firmware version refer to some hub internal package to handle the device, like dependencies. Why I think this :
- the file @nutki exposed [here](https://github.com/nathankellenicki/node-poweredup/pull/60#issuecomment-562728382)
- some videos of people who make poweredup/power function cables and control motors in train mode or linear depending on the wiring and I think power functions devices do not have firmware.

I will make some experiment to verify this.

src/devices.ts
```diff
@@ -0,0 +1,93 @@
...
+    0: "UNKNOW",
+};
+
+for (const type in devices) {
+    if (devices[+type]) {
+        const device = devices[+type];
+        deviceType[device.typeName] = device.type;
+        deviceTypeNames[device.type] = device.typeName;
+    }
+}
```
aileo: they are no longer enum and I think it is not very good.

src/devices/generic/device.ts
```diff
@@ -89,24 +97,44 @@ export class Device extends EventEmitter {
...
+    }
+
+    public get type () {
+        return this.constructor._type;
+    }
+
+    public get typeName () {
+        return this.constructor._typeName;
+    }
+
```
aileo: Get type and type name from static attributes instead of consts

src/devices/generic/device.ts
```diff
@@ -1,10 +1,17 @@
...
-import { IDeviceInterface, IDeviceMode } from "../interfaces";
+import { IDeviceInterface, IDeviceMode } from "../../interfaces";
 
-import * as Consts from "../consts";
+import { BLECharacteristic, HubType, ValueBits, ValueType } from "../../consts";
+
+export type DeviceVersion = {
+    hardware: string,
+    software: string,
+}
```
aileo: Add hardware/software version to handle modes from "ability" configuration files

src/devices/generic/device.ts
```diff
@@ -1,10 +1,17 @@
...
 
 export class Device extends EventEmitter {
+    protected static _type: number = 0;
+    protected static _typeName: string = "UNKNOW";
```
aileo: Add device type and name to static attribute to remove namespaces and consts

src/hubs/generic/basehub.ts
```diff
@@ -317,45 +306,17 @@ export class BaseHub extends EventEmitter {
...
+
         if (constructor) {
-            return new constructor(this, portId);
+            return new constructor(this, portId, versions);
         } else {
-            return new Device(this, portId, undefined, deviceType);
+            return new Device(this, portId, versions, undefined);
         }
 
     }
```
aileo: I don't know if it's a better place but now it is in the `devices.ts` collection of devices constructors

src/index-browser.ts
```diff
@@ -2,76 +2,16 @@ import * as Consts from "./consts";
...
-import { TechnicLargeLinearMotor } from "./devices/techniclargelinearmotor";
-import { TechnicMediumHubAccelerometerSensor } from "./devices/technicmediumhubaccelerometersensor";
-import { TechnicMediumHubGyroSensor } from "./devices/technicmediumhubgyrosensor";
-import { TechnicMediumHubTiltSensor } from "./devices/technicmediumhubtiltsensor";
-import { TechnicXLargeLinearMotor } from "./devices/technicxlargelinearmotor";
-import { TiltSensor } from "./devices/tiltsensor";
-import { TrainMotor } from "./devices/trainmotor";
-import { VoltageSensor } from "./devices/voltagesensor";
+import * as devices from "./devices";
+import * as hubs from "./hubs";
```
aileo: less imports in both index.

src/index-node.ts
```diff
@@ -1,76 +1,11 @@
...
 
-import { BaseHub } from "./hubs/basehub";
-import { DuploTrainBase } from "./hubs/duplotrainbase";
-import { Hub } from "./hubs/hub";
-import { MoveHub } from "./hubs/movehub";
-import { RemoteControl } from "./hubs/remotecontrol";
-import { TechnicMediumHub } from "./hubs/technicmediumhub";
-import { WeDo2SmartHub } from "./hubs/wedo2smarthub";
+export * from "./devices";
+export * from "./hubs";
```
aileo: export all hubs and all devices

src/poweredup-browser.ts
```diff
@@ -170,50 +164,33 @@ export class PoweredUP extends EventEmitter {
...
-                break;
-            case Consts.HubType.TECHNIC_MEDIUM_HUB:
-                hub = new TechnicMediumHub(device);
-                break;
-            default:
-                return;
+        const constructor = hubs[type];
+
+        if (!constructor) {
+            return;
```
aileo: Apply to hub creation the same process as the devices

src/hubs/generic/basehub.ts
```diff
@@ -1,45 +1,31 @@
...
+    public static get type () {
+        return this._type;
+    }
+    public static get typeName () {
+        return this._typeName;
+    }
+
+    protected static _type: number = 0;
+    protected static _typeName: string = "UNKNOW";
+    protected static _portMap: {[portName: string]: number} = {};
```
aileo: Add type, type name and port map to static attributes to remove namespace and related consts

aileo:
> am not sure, but I have the intuition that this firmware version refer to some hub internal package to handle the device, like dependencies. Why I think this :
> 
> * the file @nutki exposed [here](https://github.com/nathankellenicki/node-poweredup/pull/60#issuecomment-562728382) which does not mention any hub type related information
> * some videos of people who make poweredup/power function cables and control motors in train mode or linear depending on the wiring and I think power functions devices do not have firmware.
> 
> I will make some experiment to verify this.

*A few months later...*

I tested mode/device information with a single Technic Large Linear Motor on three kind of hubs, using the web bluetooth example:

| hub | hardware revision | software revision | mode 0 | mode 1 | mode 2 | mode 3 | mode 4 | mode 5 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Technic Medium Hub | 0.0.00.1000 | 0.0.00.1000 | POWER | SPEED | POS | APOS | LOAD | CALIB |
| Move Hub           | 0.0.00.0004 | 1.0.00.0000 | POWER | SPEED | POS | APOS | CALIB | STATS |
| Base Hub           | 0.0.00.1000 | 0.0.00.1000 | POWER | SPEED | POS | APOS | LOAD | CALIB |

So the **hardware** revision can change between hubs for the same device...

the mode definition is stricly identical for Technic and Base hub so I think we can rely on version number regardless of the hub type.

I don't know how to structure mode configuration: one file per device ? one per device per versions ?
Should it be json file or typescript ? Typescript is my preferred, easier to write, format/type error on compilation instead of runtime...

If we do not have matching configuration, do we load from mode messages ? If yes, how do we map mode names to events ? If no, do we fall back to the nearest ?

It is a lot of questions, but I think it's worth a try.

nathankellenicki:
_A week or so later..._

Sorry I just remembered about this. :) A few months ago I started working on a branch to do the sensor combined modes. I forgot about it until yesterday, so today I started doing more work on it. But it got me thinking. (https://github.com/nathankellenicki/node-poweredup/tree/feature/combined-sensor-modes)

A lot of it is duplication about defining how many datasets per sensor mode etc, something which as you say may be better _defined_ rather than manifested in code, perhaps through some kind of file you suggest. For this I think JSON would be better.

My worry is I don't really want to have to maintain a per-version list of modes per device, because inevitably it'll become out of date. When that happens, connections would have to become really slow as we start querying every device for their capabilities before you can use them. But I'm not really sure of a great solution for that. Toughts?

aileo:
> My worry is I don't really want to have to maintain a per-version list of modes per device, because inevitably it'll become out of date. When that happens, connections would have to become really slow as we start querying every device for their capabilities before you can use them. But I'm not really sure of a great solution for that. Thoughts?

In #60 I was reading modes from the port information messages, here are pros and cons:

Pros:
- It works whatever version the hubs/device is running
- It will simplify code for parsing values and certainly help about mode combination

Cons:
- Add latency on attach*
- Maping mode names to events

*About latency, we are talking about startup latency, not running latency so it won't affect the main usage. This does not add complexity to the lib user as the 6.x design is totally event driven so you just wait for the attach (and the attach would be fired only when modes are all good).

> It seems to be that the "main" (ie. non internal or calibration ones) are unlikely to change as they'd need to update many apps, so maybe we don't factor in versioning and just have a single file that we update?

I think lego apps are relying on the port information messages so they do not have to maintain consistent version peers. In my test above, the LOAD mode (which is a *main* sensor mode if you try to do some calibration step like for 42100) is missing on boost hub.

I can think of a compromise, using a flag we could tell the lib tu use either the internal definition or load them from port information messages:

```ts
const PoweredUP = require("node-poweredup");
const poweredUP = new PoweredUP.PoweredUP({ dynamicModes: true });
```
*Not sure about the naming...*

Some devices could just ignore this flag because of their simplicity: current, voltage, hub led, ...

aileo:
going further in #97

PR #76, opened 2020-03-03
Feature/return promises

highflying:
I found that bits of the code don't always wait for promises to return before continuing. I'm raising this PR to get your thoughts on it - I've not tested this code fully yet so don't merge it just yet!


highflying:
It've done some more testing now, so it's ready for review.

nathankellenicki:
This looks good to me - thank you for the tidy up!

PR #80, opened 2020-03-24
[feat] Power Functions IR

nathankellenicki:
This PR introduces the Power Functions IR capability of the (Boost) Color and Distance sensor.

Still to implement: Extended mode and extended channels

### DIFF
--- a/src/devices/colordistancesensor.ts
+++ b/src/devices/colordistancesensor.ts
@@ -97,7 +97,7 @@ export class ColorDistanceSensor extends Device {
             address = 1;
         }
         const message = Buffer.alloc(2);
-        // Send "Single output mode"
+        // Send "Extended toggle address command"
         message[0] = ((channel - 1) << 4) + (address << 3);
         message[1] = 6;
         return this.sendPFIRMessage(message);
--- a/src/devices/colordistancesensor.ts
+++ b/src/devices/colordistancesensor.ts
@@ -143,7 +143,7 @@ export class ColorDistanceSensor extends Device {
             address = 1;
         }
         const message = Buffer.alloc(2);
-        // Send "Combo PWD mode"
+        // Send "Combo PWM mode"
         message[0] = (((channel - 1) + 4 + (address << 3)) << 4) + this._pfPowerToPWM(powerA);
         message[1] = this._pfPowerToPWM(powerB) << 4;
         return this.sendPFIRMessage(message);
--- a/src/devices/colordistancesensor.ts
+++ b/src/devices/colordistancesensor.ts
@@ -192,7 +192,7 @@ export class ColorDistanceSensor extends Device {
 
 
     private _pfPowerToPWM (power: number) {
-        return (power < 0 ? 16 + power : power);
+        return power & 15;
     }
 
 
--- a/src/devices/colordistancesensor.ts
+++ b/src/devices/colordistancesensor.ts
@@ -145,7 +145,7 @@ export class ColorDistanceSensor extends Device {
         const message = Buffer.alloc(2);
         // Send "Combo PWD mode"
         message[0] = (((channel - 1) + 4 + (address << 3)) << 4) + this._pfPowerToPWM(powerA);
-        message[1] += this._pfPowerToPWM(powerB) << 4;
+        message[1] = this._pfPowerToPWM(powerB) << 4;
         return this.sendPFIRMessage(message);
     }
 
--- a/src/devices/colordistancesensor.ts
+++ b/src/devices/colordistancesensor.ts
@@ -175,7 +175,7 @@ export class ColorDistanceSensor extends Device {
      */
     public setColor (color: number | boolean) {
         return new Promise((resolve, reject) => {
-            if (typeof color === "boolean") {
+            if (color === false) {
                 color = 0;
             }
             if (this.isWeDo2SmartHub) {
--- a/src/devices/colordistancesensor.ts
+++ b/src/devices/colordistancesensor.ts
@@ -10,6 +10,8 @@ import * as Consts from "../consts";
  */
 export class ColorDistanceSensor extends Device {
 
+    private _pfToggleBit = 0;
+
     constructor (hub: IDeviceInterface, portId: number) {
         super(hub, portId, ModeMap, Consts.DeviceType.COLOR_DISTANCE_SENSOR);
     }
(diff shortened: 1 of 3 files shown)


src/devices/colordistancesensor.ts
```diff
@@ -10,6 +10,8 @@ import * as Consts from "../consts";
  */
 export class ColorDistanceSensor extends Device {
 
+    private _pfToggleBit = 0;
```
nutki: Unused?

src/devices/colordistancesensor.ts
```diff
@@ -79,11 +81,128 @@ export class ColorDistanceSensor extends Device {
...
+
+    /**
+     * Set the power of a Power Functions motor via IR
+     * @method ColorDistanceSensor#setPFPower
+     * @param {number} channel Channel number, between 1-4
+     * @param {string} output Outport port, "RED" (A) or "BLUE" (B)
+     * @param {number} power -7 (full reverse) to 7 (full forward). 0 is stop. 8 is brake.
+     * @returns {Promise} Resolved upon successful issuance of the command.
+     */
+    public setPFPower (channel: number, output: string, power: number) {
```
nutki:
```suggestion
    public setPFPower (channel: number, output: "RED" | "BLUE", power: number) {
```

src/devices/colordistancesensor.ts
```diff
@@ -79,11 +81,128 @@ export class ColorDistanceSensor extends Device {
...
+     * @method ColorDistanceSensor#sendPFIRMessage
+     * @param {Buffer} message 2 byte payload making up a Power Functions protocol command. NOTE: Only specify nibbles 1-3, nibble 4 should be zeroed.
+     * @returns {Promise} Resolved upon successful issuance of the command.
+     */
+    public sendPFIRMessage (message: Buffer) {
+        if (this.isWeDo2SmartHub) {
+            throw new Error("Power Functions IR is not available on the WeDo 2.0 Smart Hub");
+        } else {
+            const payload = Buffer.alloc(2);
+            payload[0] = (message[0] << 4) + (message[1] >> 4);
```
nutki: Why not generate the `message` matching the `payload` format to avoid this nibble shuffling? It seems that this is correct as far as I can tell, but it was hard for me to follow.

src/devices/colordistancesensor.ts
```diff
@@ -79,11 +81,128 @@ export class ColorDistanceSensor extends Device {
...
+     * @returns {Promise} Resolved upon successful issuance of the command.
+     */
+    public startPFMotors (channel: number, powerA: number, powerB: number) {
+        let address = 0;
+        if (channel > 4) {
+            channel -= 4;
+            address = 1;
+        }
+        const message = Buffer.alloc(2);
+        // Send "Combo PWD mode"
```
nutki:
```suggestion
        // Send "Combo PWM mode"
```

src/devices/colordistancesensor.ts
```diff
@@ -79,11 +81,128 @@ export class ColorDistanceSensor extends Device {
...
+     * @returns {Promise} Resolved upon successful issuance of the command.
+     */
+    public setPFExtendedChannel (channel: number) {
+        let address = 0;
+        if (channel >= 4) {
+            channel -= 4;
+            address = 1;
+        }
+        const message = Buffer.alloc(2);
+        // Send "Single output mode"
```
nutki:
```suggestion
        // Send "Extended toggle address command"
```

src/devices/colordistancesensor.ts
```diff
@@ -79,11 +81,128 @@ export class ColorDistanceSensor extends Device {
...
+                this.subscribe(Mode.LED);
+                this.writeDirect(0x05, Buffer.from([color]));
+            }
+            return resolve();
+        });
+    }
+
+
+    private _pfPowerToPWM (power: number) {
+        return (power < 0 ? 16 + power : power);
```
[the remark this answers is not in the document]
nathankellenicki: Thanks, good spot - I'll change this!

src/devices/colordistancesensor.ts
```diff
@@ -79,11 +81,128 @@ export class ColorDistanceSensor extends Device {
...
+    public startPFMotors (channel: number, powerA: number, powerB: number) {
+        let address = 0;
+        if (channel > 4) {
+            channel -= 4;
+            address = 1;
+        }
+        const message = Buffer.alloc(2);
+        // Send "Combo PWD mode"
+        message[0] = (((channel - 1) + 4 + (address << 3)) << 4) + this._pfPowerToPWM(powerA);
+        message[1] += this._pfPowerToPWM(powerB) << 4;
```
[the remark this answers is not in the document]
nathankellenicki: Also a good spot - copy pasta error. :)

src/devices/colordistancesensor.ts
```diff
@@ -79,11 +81,128 @@ export class ColorDistanceSensor extends Device {
...
+
+    /**
+     * Set the color of the LED on the sensor via a color value.
+     * @method ColorDistanceSensor#setColor
+     * @param {Color} color
+     * @returns {Promise} Resolved upon successful issuance of the command.
+     */
+    public setColor (color: number | boolean) {
+        return new Promise((resolve, reject) => {
+            if (typeof color === "boolean") {
```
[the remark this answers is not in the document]
nathankellenicki: The intention is to allow `setColor(false)` - since color 0 turns off the LED, I felt it was intuitive to allow a boolean false to be passed in to turn off the LED.

Passing true would cause undefined behaviour though, so I think it'd be wise for me to add a guard around this.

src/devices/colordistancesensor.ts
nathankellenicki: Yep! Will remove. 👍

src/devices/colordistancesensor.ts
nathankellenicki: Yeah, I agree it makes the code a little unclear, but my reasoning for this is to allow raw messages to be passed in in the exact order as specified in the LPF1.0 documentation. Since the hub accepts data in little endian format, the byte order doesn't match what the documentation shows, which could cause confusion to a user of this library.

It also made it easier for me to author the functions - since I could write the code to output data in the order the documentation specified, and not what the LPF2.0 hub expects.

The downside is a little bit of magical data shifting, but I think the clarity and user friendliness it provides is worth the tradeoff?

src/devices/colordistancesensor.ts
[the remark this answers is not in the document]
nutki: The intention could be more explicit with:
```
public setColor (color: number | false) {
  return new Promise((resolve, reject) => {
    if (color === false) {
      color = 0;
    }         
```
Or even (though I am a bit on the fence with the readability of it):
```
public setColor (color: number | false) {
  return new Promise((resolve, reject) => {
    color = color || 0;         
```

src/devices/colordistancesensor.ts
nutki: OK, I see your point. I did not spot that this shuffling is equivalent to shift right by one nibble to account for missing checksum followed by an endianess swap. Though while I agree that big endian makes more sense in the context of a communication protocol, note that the LPF spec itself never needs to serialize the code into bytes. That is why in my solution I went for returning a 12 bit integer instead (https://gist.github.com/nutki/e5c2987ffb897a25ef3dce21ec27b8b5)

PR #86, opened 2020-05-25
Improve web ble mailbox

aileo:
It has been a while since [this comment](https://github.com/nathankellenicki/node-poweredup/pull/64#issuecomment-566673521) but never too late...

This should improve the mailbox used in the browser while waiting for the hub type message.
It should works with multiple/partial message per frame and prevent message dropping when reading mailbox.

### DIFF
--- a/src/poweredup-browser.ts
+++ b/src/poweredup-browser.ts
@@ -132,31 +132,29 @@ export class PoweredUP extends EventEmitter {
             let buf: Buffer = Buffer.alloc(0);
             device.subscribeToCharacteristic(Consts.BLECharacteristic.LPF2_ALL, (data: Buffer) => {
                 buf = Buffer.concat([buf, data]);
-                const len = buf[0];
-                if (len >= buf.length) {
+                while (buf[0] <= buf.length) {
+                    const len = buf[0];
                     const message = buf.slice(0, len);
                     buf = buf.slice(len);
                     if (message[2] === 0x01 && message[3] === 0x0b) {
-                        process.nextTick(() => {
-                            switch (message[5]) {
-                                case Consts.BLEManufacturerData.REMOTE_CONTROL_ID:
-                                    resolve(Consts.HubType.REMOTE_CONTROL);
-                                    break;
-                                case Consts.BLEManufacturerData.MOVE_HUB_ID:
-                                    resolve(Consts.HubType.MOVE_HUB);
-                                    break;
-                                case Consts.BLEManufacturerData.HUB_ID:
-                                    resolve(Consts.HubType.HUB);
-                                    break;
-                                case Consts.BLEManufacturerData.DUPLO_TRAIN_BASE_ID:
-                                    resolve(Consts.HubType.DUPLO_TRAIN_BASE);
-                                    break;
-                                case Consts.BLEManufacturerData.TECHNIC_MEDIUM_HUB:
-                                    resolve(Consts.HubType.TECHNIC_MEDIUM_HUB);
-                                    break;
-                            }
-                            debug("Hub type determined");
-                        });
+                        switch (message[5]) {
+                            case Consts.BLEManufacturerData.REMOTE_CONTROL_ID:
+                                resolve(Consts.HubType.REMOTE_CONTROL);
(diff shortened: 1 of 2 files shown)


src/poweredup-browser.ts
```diff
@@ -132,31 +132,29 @@ export class PoweredUP extends EventEmitter {
             let buf: Buffer = Buffer.alloc(0);
             device.subscribeToCharacteristic(Consts.BLECharacteristic.LPF2_ALL, (data: Buffer) => {
                 buf = Buffer.concat([buf, data]);
-                const len = buf[0];
-                if (len >= buf.length) {
```
aileo, line 136: I may be wrong but this condition does not look good to me: expected length greater than buffer size ?

src/poweredup-browser.ts
```diff
@@ -132,31 +132,29 @@ export class PoweredUP extends EventEmitter {
...
+                while (buf[0] <= buf.length) {
+                    const len = buf[0];
```
aileo, line 136: Loop for all available messages

src/poweredup-browser.ts
```diff
@@ -132,31 +132,29 @@ export class PoweredUP extends EventEmitter {
...
                     const message = buf.slice(0, len);
                     buf = buf.slice(len);
                     if (message[2] === 0x01 && message[3] === 0x0b) {
-                        process.nextTick(() => {
```
aileo, line 140: This is already in an event callback, I don't see the point of the `process.nextTick()`

src/webbleabstraction.ts
```diff
@@ -99,11 +99,14 @@ export class WebBLEDevice extends EventEmitter implements IBLEAbstraction {
             return callback(buf);
         };
         this._characteristics[uuid].addEventListener("characteristicvaluechanged", this._listeners[uuid]);
-        for (const data of this._mailbox) {
+
+        const mailbox = Array.from(this._mailbox);
+        this._mailbox = [];
```
aileo, line 104: copy mailbox content and reset mailbox. It prevent message being dropped if callback is "posting" to the mailbox again.

src/poweredup-browser.ts
nathankellenicki: BLE has QoS to scale MTU size depending on network conditions. It's rare, but under certain poor/highly contested conditions the sender may not be able to fit the whole message into a single packet, meaning we get partial data instead of a full message.

The first byte of the message we receive tells us the length of data we should expect. If we don't have all the data yet, we skip parsing, concatenating data until we get a full message (line 134).

src/poweredup-browser.ts
nathankellenicki: My memory is fading me on this one, but I think this was in place as advertisement packets (with manufacturer data to determine hub type) came in slightly after the discovery, and this was in place to allow it to be processed already. If I recall correctly this was particularly a problem with the Boost Move Hub.

It may be it's not required since the v6 rearchitecting, but I'll need to do some solid testing with different hub types before I feel comfortable removing it.

src/poweredup-browser.ts
aileo: > I think the length check needs to be reintroduced, but lgtm otherwise. :) I need to spend a bit of time testing to ensure the removal of the `process.nextTick` doesn't cause any issues.

The length check is still there but in a while to process as many messages as possible.
My comment was about the sign of the comparison in the if: it seems to parse messages if the buffer is equal or **smaller** than the advertised message length.

I will do some more tests tomorrow to check with the boost hub.

nathankellenicki: lawl, good catch - the comparison is turned around. Turns out I shouldn't be allowed anywhere near a text editor. :D

aileo:
> I think the length check needs to be reintroduced, but lgtm otherwise. :) I need to spend a bit of time testing to ensure the removal of the `process.nextTick` doesn't cause any issues.

Tested Boost, Control+ and PUP hub (separatly and together) using the Web Bluetooth example and everything seems ok.

The `nextTick` does not have a significant cost, it could be removed from this PR.

PR #99, opened 2020-06-28
Fix message length comparison

aileo:
I think the expression was wrong: we want to parse if the message length is less or equal to buffer size.


nathankellenicki:
@aileo Good catch! I guess this only ever worked because we've never had a message split over multiple packets, and always parsed them one message at a time. Luckilly! :)

aileo:
> I guess this only ever worked because we've never had a message split over multiple packets, and always parsed them one message at a time.

I wonder if this could be the root cause of #63, we will probably never know

PR #111, opened 2020-10-16
Replace magic numbers with constants

jncraton:
This replaces several magic numbers in the lpf2hub implementation with constants that were already present in src/consts.ts. This should enhance readability and maintainability. Comments that were made redundant by including the constant names have been removed.

There are still quite a few magic numbers in here that could be replaced, particularly in the byte arrays used to build messages, but I didn't attempt to clean those up.


nathankellenicki:
Thanks for the contribution, looks great! As new magic numbers are spotted, they can migrated over. This is a great start.

PR #114, opened 2020-12-02
Re-connection to already known hubs

pwimmer:
This pull requests improves the Web Bluetooth support. The _scan_ method scans for advertisements of hubs that have been connected before. The old scan method that caused a popup in the browser has been renamed to _requestDevice_.

I've tested with Chrome 87. The following flags are required:

- chrome://flags/#enable-experimental-web-platform-features
- chrome://flags/#enable-web-bluetooth-new-permissions-backend

There are still issues that the LED color does not change in the sample application, but I had the same issue with the original code.

### DIFF
--- a/examples/web_bluetooth.html
+++ b/examples/web_bluetooth.html
@@ -57,8 +57,8 @@ setInterval(() => {
     const hubs = poweredUP.getHubs(); // Get an array of all connected hubs
     document.getElementById("color").style.backgroundColor = PoweredUP.Consts.ColorNames[color];
     hubs.forEach(async (hub) => {
-        const led = await hub.waitForDeviceByType(PoweredUP.Consts.DeviceType.HUB_LED);
-        led.setColor(color); // Set the color
+         const led = await hub.waitForDeviceByType(PoweredUP.Consts.DeviceType.HUB_LED);
+         led.setColor(color); // Set the color
     })
     color++;
     if (color > 10) {
@@ -82,7 +82,23 @@ const renderHub = function (hub) {
 
 const scan = function () {
     if (PoweredUP.isWebBluetooth) {
-        poweredUP.scan(); // Start scanning for hubs
+        poweredUP.scan(); // Start scanning for known hubs
+    } else {
+        alert("Your browser does not support the Web Bluetooth specification.");
+    }
+}
+
+const stop = function () {
+    if (PoweredUP.isWebBluetooth) {
+        poweredUP.stop(); // Stop scanning for known hubs
+    } else {
+        alert("Your browser does not support the Web Bluetooth specification.");
+    }
+}
+
+const requestDevice = function () {
+    if (PoweredUP.isWebBluetooth) {
+        poweredUP.requestDevice(); // Start scanning for new hubs
     } else {
         alert("Your browser does not support the Web Bluetooth specification.");
     }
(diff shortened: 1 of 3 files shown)


aileo:
That is a nice improvement.
Instead of changing the behavior of the scan method (which would break backward compatibility), do you think it could be a parameter of the scan method ?

pwimmer:
> Instead of changing the behavior of the scan method (which would break backward compatibility), do you think it could be a parameter of the scan method ?

The reason why I renamed the _scan_ method is that the new scan/stop methods act very similar to the Node implementation so I thought they should use the same names. The _requestDevice_ method has no equivalent in Node so I thought it should have a new name.

As _scan_ and _requestDevice_ do different things it is cleaner to have separate methods. But I'm fine with a parameter if backward compatibility must be preserved.

aileo:
> > Instead of changing the behavior of the scan method (which would break backward compatibility), do you think it could be a parameter of the scan method ?
> 
> The reason why I renamed the _scan_ method is that the new scan/stop methods act very similar to the Node implementation so I thought they should use the same names. The _requestDevice_ method has no equivalent in Node so I thought it should have a new name.

Seems legit.

This lib uses the term "Device" to refer to items connected to hubs (motors, sensors...), I think `requestDevice` could be confusing, maybe `requestHub` could be more accurate.

nathankellenicki:
@pwimmer I'm fine with breaking changes when good rationale is presented, and this is good rationale. :) Synchronicity between the Node.js and Web Bluetooth API is certainly a plus to reduce the amount of mental hoops would need to jump over when working between the two.

@aileo 's suggestion for `requestHub` over `requestDevice` is a good one to maintain naming conventions with the rest of the lib.

This is a great additional piece of functionality, many thanks for the contribution! My main concern is the necessity for end users to enable experimental features in the browser - one of the tenets I tried to follow is to make it so that the lib could just be published to a website which end users could visit with no further barriers. Do you have additional information on when these features may land proper? A link to some docs or discussions perhaps?

Thanks, and happy holidays to both. :)

examples/web_bluetooth.html
```diff
@@ -82,7 +82,23 @@
 
 const scan = function () {
     if (PoweredUP.isWebBluetooth) {
-        poweredUP.scan(); // Start scanning for hubs
+        poweredUP.scan(); // Start scanning for known hubs
+    } else {
+        alert("Your browser does not support the Web Bluetooth specification.");
```
nathankellenicki, line 87: Could we have this throw instead of alert? Having it throw gives users of this library the ability to catch the error and present whatever UX they see as best for their use case, instead of mandating an alert.

examples/web_bluetooth.html
```diff
@@ -82,7 +82,23 @@
...
+    }
+}
+
+const stop = function () {
+    if (PoweredUP.isWebBluetooth) {
+        poweredUP.stop(); // Stop scanning for known hubs
+    } else {
+        alert("Your browser does not support the Web Bluetooth specification.");
```
nathankellenicki, line 95: Same here - could this throw instead.

examples/web_bluetooth.html
pwimmer: This is the sample application, the alert has no impact on the users of the library.

nathankellenicki: You're absolutely right, I didn't spot the context. Apologies!

nathankellenicki:
@pwimmer I got around to testing this PR properly tonight. Unfortunately it looks like these features still haven't landed as they need to be turned on in Chrome flags.

In addition, I wasn't able to get this reliably working. After pairing, it wouldn't automatically reconnect to any previously paired hubs until I opened the pairing dialog again, at which point it would automatically connect, even without selecting the device and pairing. It seems unfortunately there are still some bugs in the implementation.

pwimmer:
> In addition, I wasn't able to get this reliably working. After pairing, it wouldn't automatically reconnect to any previously paired hubs until I opened the pairing dialog again, at which point it would automatically connect, even without selecting the device and pairing. It seems unfortunately there are still some bugs in the implementation.

Automatic reconnection requires a https server. It doesn't work with http.

Issue #118, opened 2020-12-27
Self-compiled library never returns promise


Debenben:
Using `const PoweredUP = require("node-poweredup");` the function `motorA.rotateByDegrees(20,10).then(function(value) {console.log("test");});` prints "test" after successful completion as expected. 

If I compile node-poweredup myself, everything works as expected except for the promises like above, which never return. To compile I do the following:

```
git clone https://github.com/nathankellenicki/node-poweredup.git
cd node-poweredup
npm install
npm run build
```

When I replace `const PoweredUP = require("node-poweredup");` with `const PoweredUP = require(".");` it behaves differently and the promise never returns and nothing gets printed.

I was not sure if my problem was in relation to this library so I already posted a question here: https://stackoverflow.com/questions/65470276/javascript-promise-never-resolves/65471208#65471208

nathankellenicki:
There's quite a bit that could be going wrong here, it's a bit difficult to know where to start.

> When I replace const PoweredUP = require("node-poweredup"); with const PoweredUP = require("."); it behaves differently and the promise never returns and nothing gets printed.

Where is this file located? Inside the `node-poweredup` directory? Or somewhere else?

You also mention you receive the following in your linked Stackoverflow question:

> error TS2794: Expected 1 arguments, but got 0. Did you forget to include 'void' in your type argument to 'Promise'?

What is your host application? Without knowing what you're trying to write (I think saw you mention Electron?) it's difficult, but assuming it's Typescript, there's a version mismatch there. Typescript 4.1 made the previously required argument to `resolve` explicitly required (instead of optional), so it will throw unless you declare the return type of the promise as void when resolving with no parameters.

You can read about it here: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#resolves-parameters-are-no-longer-optional-in-promises

A few days ago I myself made the process of bringing these promises up to speed, so I also suspect you're using a slightly older version of this library. I recommend updating to the latest which explicitly sets the return type of these promises to void (including the one you mention in your Stackoverflow question).

Debenben:
Thank you a lot for this fast, detailed answer. 

The application I am trying to write is using Electron and Electron-rebuild. It is my first time ever writing something in JavaScript or Typescript, still work in progress and was not intended for publication. I uploaded it anyway so you can have a look: https://github.com/Debenben/babyele. The version of Typescript there seems to be 4.1.3, I have to test what happens if I change it.

[...]

PR #119, opened 2020-12-29
fix issue #118 (partially)

Debenben:
fix wrong constant, should be 0x82 like in previous version.


nathankellenicki:
Thanks, good spot!

nathankellenicki:
Published in `v6.9.1`

PR #130, opened 2021-05-09
Rename events enum from AlertPayload to Event

sciguy16:
Looks like the Event enum got the wrong name, so I've updated it


nathankellenicki:
Thanks!

PR #135, opened 2021-06-08
Fix path for constants after renaming it

10on:
After merging PR https://github.com/nathankellenicki/node-poweredup/pull/130 project can not build from master branch. This fix should be resolve this issue.


nathankellenicki:
Thanks for this! Not sure why it continued building for me, but I can see the error. Merged!

PR #143, opened 2022-01-09
update enum documentation

Debenben:


nathankellenicki:
Wow, how did I let it get this out of sync. Thanks!

PR #145, opened 2022-01-09
change device mode after feedback only

Debenben:
During a couple of different tests, I came across the following problem, here illustrated in a reproducible way for motor rotation:

`
    const motorA = await hub.waitForDeviceAtPort("A");
    motorA.requestUpdate(); // requestUpdate for default mode 0x00
    motorA.on("rotate", () => { // motorA._mode is changed to 0x02
        console.log("recieved rotate");
    });
`


nathankellenicki:
Thanks for this. I need to do a bit more testing with this so I'll do some this week with different hub types. I recognise the original issue though, good find.

Debenben:
@nathankellenicki Did you encounter any issues with this? For me this change works well and was sufficient to solve all issues.

If you are unsure, one could add a mechanism with a timeout to request the mode again if no response is recieved and additionally ignore messages that do not have the expected size.

nathankellenicki:
18 months later and merged. :D Thanks again!

PR #149, opened 2022-01-09
change colordistancesensor distance calculation

Debenben:
I know the distances in general are not very accurate and it depends on the material used for testing how the infrared based distance value and light based partial value combine. But still, with this simplification my overall results were better.

### DIFF
--- a/src/devices/colordistancesensor.ts
+++ b/src/devices/colordistancesensor.ts
@@ -63,10 +63,10 @@ export class ColorDistanceSensor extends Device {
                 const partial = message[7];
 
                 if (partial > 0) {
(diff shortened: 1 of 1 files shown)


nathankellenicki:
This one is tricky. I originally had this set to -22 a couple of years ago, but I discovered that the delta varied wildly depending on the color of the object as well as the ambient light in the room. Testing it now using my hand, a white object appears somewhat accurate, whereas my hand is approx 20mm off.

Saying that, I've been considering changing it for a little bit based on my experience lately with a Christmas train project, so I'll accept this back for now (and change the WeDo 2.0 sensor too). We'll see if there's any complaints. :)

Issue #157, opened 2022-05-27
Race in _determineLPF2HubType


antlai:
In Android smartphones  (with webbluetooth) device.subscribeToCharacteristic() and device.writeToCharacteristic() cannot run concurrently without triggering the error in  #131 , i.e., GATT operation failed for unknown reason.

To avoid that, in _determineLPF2HubType we need to await device.subscribeToCharacteristic() before starting  device.writeToCharacteristic(). We also need subscribeToCharacteristic() to return the promise returned by startNotifications() (in webbleabstraction.ts)

This seems to have fixed all my connection problems.

It is just four lines changed, I'll submit a pull request...

Thanks for a great library!

Antonio

vene4ka:
not helping with #131

Issue #131, opened 2021-05-11
Uncaught (in promise) DOMException: GATT operation failed for unknown reason.


DanieleBenedettelli:
Hi.
I am using your library by adding it like so:
<script src="https://cdn.jsdelivr.net/npm/node-poweredup@latest/dist/browser/poweredup.js"></script>

Then I scan for LEGO BOOST hubs like so:
```javascript
if (PoweredUP.isWebBluetooth) {
        poweredUP.scan(); // Start scanning for hubs
} else {
    alert("Your browser does not support the Web Bluetooth specification. Maybe you should just forgot https://");
}
```

Then I use the callback
```javascript
poweredUP.on("discover", async function(hub) { // Wait to discover hubs
	console.log(" poweredUP on discover");
	await hub.connect(); // Connect to hub
[...]
}
```
Inside there, I use other handlers:
hub.on("attach", function(device) {
hub.on("disconnect", function() {

I tested my app successfully on Win 10 laptops with Chrome, but on smartphones connecting gives the error:
Uncaught (in promise) DOMException: GATT operation failed for unknown reason.

I had a lot of troubles installing VS to modify the library, otherwise I would have suggested a solution. Maybe, maybe, it's a matter of timing in data exchange, but it's just my hard guess.

Thanks!

PR #158, opened 2022-05-27
Fix for #157 (and possibly #131)

antlai:
Fix race in _determineLPF2HubType that makes a connection fall in Android web Bluetooth (see #157 )


Debenben (approved):
I did not have any connection problems so far, but the change looks good and does not break anything for me. Thanks!

PR #159, opened 2022-05-29
rework port output command handling

Debenben:
Initially I only planned to do some small improvements to prevent my port input commands from clogging up such that they were already outdated when being transmitted. But the whole undertaking got more and more inconsistent and messy, so I ended up doing a large-scale rework for all port output commands.

A large majority of code using the library should be compatible with the changes, but some things would change. Rough overview of most important breaking change without paying attention to detail:

**same behavior:**
`motorA.rotateByDegrees(30).then(motorA.rotateByDegrees(30));`
Result: motorA rotates by 60 degrees in total

`await motorA.rotateByDegrees(30);
 await motorA.rotateByDegrees(30);`
Result: motorA rotates by 60 degrees in total

`motorA.rotateByDegrees(30);
 motorB.rotateByDegrees(30);`
Result: motorA and motorB start rotating at the same time by 30 degrees each

`motorA.setAccelerationTime(300);
 motorA.rotateByDegrees(30);`
Result: the acceleration time is set to 300 ms and then motorA rotates by 30 degrees

**different behavior:**
`motorA.rotateByDegrees(30);
 motorA.rotateByDegrees(30);`
Result old: motorA rotates by 30 degrees
Result new: motorA rotates by 60 degrees

**new feature:**
`motorA.rotateByDegrees(90);
 motorA.rotateByDegrees(30, true);`
Result new: motorA rotates by 30 degrees. 

I tested with a technic hub and two motors so far it works great for me and I did not encounter any problems, therefore I decided to put this change up for discussion even though not every device has been tested.

Commit message:
use explicit queues for port output commands;
this helps the device to remain responsive by preventing old port
output commands from clogging the bluetooth transmission.

add improved interrupt property;
if true, ALL previous commands are discarded
if false, NONE of the previous commands are discarded
this changes the previous behavior where commands queued for
transmission could not be discarded and transmitted commands
would not wait for the device to complete the previous command.

send commands to queue by default (except break and stop);
this keeps and improves the ability to accept commands faster than
they can be transmitted via bluetooth and ensures that no commands
discard previous commands in case they are not explicitly chained.

provide feedback about success of commands.

### DIFF
--- a/src/consts.ts
+++ b/src/consts.ts
@@ -703,3 +703,27 @@ export enum MarioColor {
     BROWN = 0x6a00,
     CYAN = 0x4201,
 }
+
+/**
+ * @typedef CommandFeedback
+ * @param {number} TRANSMISSION_PENDING 0x00 waiting for previous comands to complete transmission or execution
+ * @param {number} TRANSMISSION_BUSY 0x10 waiting for device to acknowledge reception
+ * @param {number} TRANSMISSION_DISCARDED 0x44 other command for immediate execution has been recieved or device disconnected
+ * @param {number} EXECUTION_PENDING 0x20 device is waiting for previous command to complete
+ * @param {number} EXECUTION_BUSY 0x21 device is executing the command
+ * @param {number} EXECUTION_DISCARDED 0x24 device discarded the command e.g. due to other command for immediate execution
+ * @param {number} EXECUTION_COMPLETED 0x22 device reported successful completion of command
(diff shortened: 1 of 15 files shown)


Debenben:
@nathankellenicki Thank you for merging some pull requests! For me this mechanism has been working well, I think it would be a benefit for the official library, too. I rebased the pull request, let me know if you have some questions, suggestions or would like me to do some changes.

PR #171, opened 2022-10-08
Extend function to manually detach a device

Debenben:
Change of the undocumented manuallyAttachDevice function to allow detaching a device. Change does not impact normal usage of library.


nathankellenicki:
`manuallyDetachDevice` implemented in v10.1.0, closing

